PR #19, opened 2018-10-10
update README.md

Behnam-Emamian:


xuzhg:
@Behnam-Emamian Nice catch. Thanks.

Issue #21, opened 2018-10-11
$select values should be comma-separated values.


Behnam-Emamian:
$select values should be comma-separated values. maybe for parameters like $expand and ...

https://swagger.io/docs/specification/serialization/

we need to set "style": "simple"

xuzhg:
Thanks for your input and contribution.

PR #22, opened 2018-10-12
set $select, $orderby and #expand as comma-separated values

Behnam-Emamian:
See related issue #21.


xuzhg:
@Behnam-Emamian I approved the changes. But, maybe we should update some test cases. Don't worry, i will update them if have.

PR #24, opened 2018-10-13
fix some misspellings in README.md

umalog:


xuzhg:
@umalog Thanks for your catching.

PR #30, opened 2019-05-08
Fix Action Response Serialization

mispeer:
Currently, all action responses are being generated as 204. However, per the OASIS Mapping documentation 4.5.3 :
HTTP response code [of an action should be ] 204 if the response has no body
...200 for other success codes...

For a specific example, the following xml:
<Action Name="GetPeersForTrip" IsBound="true">
        <Parameter Name="personInstance" Type="Microsoft.OData.Service.Sample.TrippinInMemory.Models.Person" />
        <Parameter Name="userName" Type="Edm.String" Nullable="false" Unicode="false" />
        <Parameter Name="tripId" Type="Edm.Int32" Nullable="false" />
        <ReturnType Type="Collection(Microsoft.OData.Service.Sample.TrippinInMemory.Models.Person)" />
      </Action>

Should generate a 200 response. 

Modified UnitTest to pass expected responseCode as InlineData


xuzhg:
Merged. Thanks.

PR #53, opened 2020-03-17
Change all query parameters from "simple" to "form"

glombek:
"simple" is not supported in query parameters for OpenAPI v3. https://swagger.io/docs/specification/describing-parameters/


darrelmiller:
@glombek Thank you for identifying, this and resolving the issue.  As the person who actually wrote that part of the OpenAPI V3 spec, first I'd like to apologize for how convoluted it is.  It was very challenging to try and respect the rules of RFC6570 and keep compatibility with OpenAPI and deal with its constraints.  I spent some time reviewing the change you made and the spec document and setting the style to form and explode to true definitely seems to be the right way to describe the OData parameters that are comma delimited arrays of values.

PR #61, opened 2020-05-13
Added simple PathPrefix to ConvertSettings to match ODataRoute prefix without affecting 'servers' declaration in OAS file.

### LINKED ISSUE #60: Paths can't be generated with a routePrefix.
It is pretty common to use a rout prefix such as "odata" or "api" for paths in an OData API.
Adding this into the generated paths would create an OAS doc that more accurately matches the implementation.

### Assemblies affected

Affects Microsoft.OpenAPI.OData.Reader.

### Steps to reproduce

Map a route with a call like the following:
  endpoints.MapODataRoute("odata", "api", this.GetEdmModel());

### Expected result
Paths are along the lines of
/api/entitySet1
/api/entitySet2

### Actual result
Paths are along the lines of
/entitySet1
/entitySet2

### Additional detail
I'm proposing to add a pathPrefix string to the Settings class, and simply include that as a prefix to all generated paths.  That provides an easy way to match the routePrefix with the actual paths.

garethj-msft:
Fixes #60 

Added simple PathPrefix to ConvertSettings to match ODataRoute prefix without affecting 'servers' declaration in OAS file.

Add test to match.

This aids with mocking tools that only understand absolute paths and cannot parse server entries to split DNS segments from path segments.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPath.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPath.cs
@@ -146,6 +146,13 @@ namespace Microsoft.OpenApi.OData.Edm
 
             HashSet<string> parameters = new HashSet<string>();
             StringBuilder sb = new StringBuilder();
+
+            if (!string.IsNullOrWhiteSpace(settings.PathPrefix))
+            {
+                sb.Append("/");
+                sb.Append(settings.PathPrefix);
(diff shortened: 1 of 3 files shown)


xuzhg:
Looks good to me.

garethj-msft:
Thanks Sam - pls merge when you're happy!  Love this tool.

PR #72, opened 2020-07-07
Fixes OpenAPI Links and makes them optional

### LINKED ISSUE #64: Links tags should be optional.
Links tags should be optional in the output, driven by a bool in the settings object.
I don't always want links tags in my OAS document..

As a side note, the links tags generated are currently completely broken, so this is a key reason for turning them off, however, even when they are working correctly, I'd like to choose whether to have them or not.

irvinesunday:
Ths PR fixes #64 

Proposes:
- Fixing the `operationIds` and `parameters` of generated OpenAPI Links.
- Making optional the generation of OpenAPI Links through a `boolean` setting.
- Updating the `Multiple.Schema.OpenApi.json` and `Multiple.Schema.OpenApi.yaml` test docs. with sample OpenAPI Link tags and content and updating their respective tests appropriately by enabling Links generation.
- Removing the OpenAPI Links tags from the other test docs.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs
@@ -9,6 +9,7 @@ using Microsoft.OpenApi.OData.Common;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.Any;
 using Microsoft.OpenApi.OData.Edm;
+using System.Linq;
 
 namespace Microsoft.OpenApi.OData.Generator
 {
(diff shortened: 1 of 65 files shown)


src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
```diff
@@ -44,15 +44,15 @@ public virtual OpenApiOperation CreateOperation(ODataContext context, ODataPath
             // Security
             SetSecurity(operation);
 
+            // Parameters
```
xuzhg, line 47: The place of function matters here?

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs
```diff
@@ -22,32 +23,83 @@ internal static class OpenApiLinkGenerator
...
+            bool targetMultiplicity = false)
         {
-            Utils.CheckArgumentNull(context, nameof(context));
-            Utils.CheckArgumentNull(entityType, nameof(entityType));
-            Utils.CheckArgumentNullOrEmpty(sourceElementName, nameof(sourceElementName));
-
             IDictionary<string, OpenApiLink> links = new Dictionary<string, OpenApiLink>();
-            foreach (IEdmNavigationProperty np in entityType.DeclaredNavigationProperties())
+
+            if (!targetMultiplicity)
```
xuzhg, line 39: >if (!targetMultiplicity) [](start = 12, length = 24)

if targetMultiplicity == true, this function seems doing nothing?
Why don't you check this boolean out of this function call?

If (!targetMultiplicity)
{
      var links  = context.CreateLinks(...);
}

src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
irvinesunday: Yes. It needs to be processed before `SetResponses()` so that we can use the parameters in the argument [here](https://github.com/irvinesunday/OpenAPI.NET.OData/blob/a72873cf6170c49785c82f4a6ad29c1757376aef/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyGetOperationHandler.cs#L97) when calling the `CreateLinks()` method; which will get used [here](https://github.com/irvinesunday/OpenAPI.NET.OData/blob/a72873cf6170c49785c82f4a6ad29c1757376aef/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs#L50-L57).

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs
irvinesunday: Good catch. Resolved.

src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
xuzhg, line 47: >// Parameters [](start = 12, length = 13)

Please add a detail comments here to maintenance. For example, we must call SetParameters ahead of ...., and Why?

irvinesunday: Added helpful comment.
Resolved

PR #88, opened 2021-02-09
Fixed paths when Operation is bound to a type derived from the type of a nav prop.

garethj-msft:
Fixed paths when Operation is bound to a type derived from the type of a nav prop.

Added a setting to make this (and navSource inheritance too) require DerivedTypeConstraint in the model, as it can produce a lot of extra paths if nav props are very weakly typed.

Added tests to match.


xuzhg (approved):

src/Microsoft.OpenApi.OData.Reader/Edm/IODataPathProvider.cs
```diff
@@ -24,7 +24,8 @@ public interface IODataPathProvider
         /// Generate the list of <see cref="ODataPath"/> based on the given <see cref="IEdmModel"/>.
         /// </summary>
         /// <param name="model">The Edm model.</param>
+        /// <param name="settings">The conversion settings.</param>
         /// <returns>The collection of built <see cref="ODataPath"/>.</returns>
-        IEnumerable<ODataPath> GetPaths(IEdmModel model);
+        IEnumerable<ODataPath> GetPaths(IEdmModel model, OpenApiConvertSettings settings);
```
xuzhg, line 29: This is a breaking change. But, I am ok to add this.

xuzhg:
Merged at e1763727f5d3bf9496c3258d325d2db76e96185a

PR #89, opened 2021-03-15
Expand containment navigation properties

irvinesunday:
This PR proposes:
- Expanding all containment navigation properties.
- Update the `Graph.Beta.OData.xml` CSDL test file by removing the containment for the nav. prop. `workbook` as its containment level is deeply nested, blowing up the resulting OpenAPI doc.
- Updating the `ODataPathProvider` test to account for the new paths count.
- Updating the generation of media stream paths for nav. props to be executed before expansion of the nav. prop.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+﻿[*.{cs,vb}]
+
+# IDE0009: Member access should be qualified.
+dotnet_diagnostic.IDE0009.severity = none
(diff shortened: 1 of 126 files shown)


irvinesunday:
Closing this as there's a newer and cleaner PR: https://github.com/microsoft/OpenAPI.NET.OData/pull/94

Issue #91, opened 2021-04-06
Parameters should not have both example and examples properties


irvinesunday:
Generated OpenAPI doc has both _example_ and _examples_ properties in the parameter section of some operations. This is invalid as they are mutually exclusive. This results in an invalid doc. 
See OpenAPI documentation: https://swagger.io/docs/specification/adding-examples/

### Assemblies affected

1.0.6

### Steps to reproduce

Generate the OpenAPI doc. from the CSDL: https://graph.microsoft.com/v1.0/$metadata
Check the path under operationId: **groups.group.GetGroup**

### Expected result
Within this path, under parameters, there should only be one occurrence of either _example_ or _examples_ 

### Actual result
There are two occurrences of _example_ and _examples_

![image](https://user-images.githubusercontent.com/40403681/113774060-765d1180-972f-11eb-8858-ef6a71898e0b.png)

### Additional detail

Related to: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/511

PR #92, opened 2021-04-06
Fix: Removes example property in path parameter objects

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/91
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/511

- The current _Example_ property defined under parameters is currently only used to define the _DocumentationURL_ property value. This can be appended to the _Description_ property. This will help in preventing the potential occurrences of both _example_ and _examples_ properties, which are mutually exclusive. 
- Tests have been added to validate this fix.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
@@ -187,8 +187,6 @@ namespace Microsoft.OpenApi.OData.Operation
             foreach (var param in customParameters)
             {
                 string documentationUrl = null;
-                string paramDescription;
-
                 if (param.DocumentationURL != null)
                 {
                     documentationUrl = $" Documentation URL: {param.DocumentationURL}";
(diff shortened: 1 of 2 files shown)


src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
```diff
@@ -186,23 +186,37 @@ protected static void AppendCustomParameters(OpenApiOperation operation, IList<C
...
+                // DocumentationURL value is to be appended to
+                // the parameter Description property
+                if (param.Description == null)
+                {
+                    paramDescription = documentationUrl?.Remove(0, 1);
+                }
+                else
+                {
+                    paramDescription = param.Description + documentationUrl;
+                }
```
thewahome, line 206: Changing the if statements here helps someone to read the code more quickly. It also moves the declaration closer to its assignment

```suggestion
               string paramDescription = (param.Description == null) ? documentationUrl?.Remove(0, 1) : param.Description + documentationUrl;
```

xuzhg (approved):

peombwa (approved):

Issue #93, opened 2021-04-07
Expand all containment navigation properties


irvinesunday:
Some containment navigation properties aren't being expanded.

### Assemblies affected

1.0.6

### Steps to reproduce

Generate the OpenAPI doc. from the CSDL: https://graph.microsoft.com/v1.0/$metadata
Notice example missing path `/users/{user-id}/drive/items/{parent-item-id}/children` 

![image](https://user-images.githubusercontent.com/40403681/113844408-1bfa9a00-979d-11eb-8994-324722023c64.png)
![image](https://user-images.githubusercontent.com/40403681/113844663-64b25300-979d-11eb-8f59-b975b57d8607.png)

In the above scenario, we are not expanding on the navigation properties under the `drive` entity type. We only get up to the path `/users/{user-id}/drive`. In this case, `drive` is a containment navigation property and should be expanded.

### Expected result

We should be able to navigate to this ex. path: `/users/{user-id}/drive/items/{parent-item-id}/children`

### Actual result

We can only navigate to this ex. path: `/users/{user-id}/drive`

### Additional detail

Currently, we are only expanding containment navigation properties whose entity types don't define any navigation source in the EntityContainer.

Related to https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/393 and https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/129

PR #94, opened 2021-04-07
Fix: Expand containment navigation properties

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/93
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/393
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/559

This PR proposes:
- Checking the _Navigability_ and _IndexabilityByKey_ of navigation properties before creating navigation property paths.
- Expanding all containment navigation properties.
- Update the `Graph.Beta.OData.xml` CSDL test file:
  -  adds `Navigability: None` `NavigationRestrictions` annotation to `driveItem/workbook`. This navigation should not be navigable. 
  - Adds a `joinedGroup` collection navigation property to entity type `user` and a subsequent `IndexableByKey: false` `NavigationRestrictions` for testing purposes.
- Updating the `ODataPathProviderTests.cs`:
  - to account for the new path count generated from `Graph.Beta.OData.xml` CSDL to OpenAPI conversion.
  - to validate that we are evaluating _Navigability_ and _IndexabilityByKey_ for navigation properties before generating navigation property paths.
- Updating the generation of media stream paths for nav. props to be executed before expansion of the navigation properties.


test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
```diff
@@ -48,7 +48,7 @@ public void GetPathsForGraphBetaModelReturnsAllPaths()
 
             // Assert
             Assert.NotNull(paths);
-            Assert.Equal(4887, paths.Count());
+            Assert.Equal(17313, paths.Count());
```
xuzhg, line 51: The path number increases almost 5 times the original number?

xuzhg (approved):

test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
irvinesunday: Because of all the valid containment navigation properties that are now being expanded.

PR #100, opened 2021-04-19
Add support for EdmTypeKind.TypeDefinition

jukkahyv:
We have an OData model with TypeDefinition.

```
<TypeDefinition Name="MyID" UnderlyingType="Edm.Int64" />
```

Currently this crashes with `OpenApiException` ("Not supported") in `OpenApiExampleGenerator.GetTypeNameForExample`.

I made a change handling `EdmTypeKind.TypeDefinition` by extracting the underlying (primitive) type. I'm not 100% sure if this is the correct way of handling it, but at least it now doesn't crash and generates valid OpenAPI schema.

Please let me know if I should write tests or make some other changes.


xuzhg (approved):

xuzhg:
@jukkahyv Thanks for your contribution.

Issue #103, opened 2021-04-27
Output OpenAPI path operations descriptions from vocabulary annotations


irvinesunday:
There is need to populate the OpenAPI path operations descriptions from the descriptions vocabulary annotations from the input CSDL doc. Currently, this is not supported.

### Assemblies affected

1.0.7

### Steps to reproduce

Generate an OpenAPI doc. from the CSDL: https://graph.microsoft.com/v1.0/$metadata.

### Expected result

We should have a description property in path operations.

### Actual result

No description property surfaced in path operations.

### Additional detail

Related to: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/529

PR #104, opened 2021-04-27
Describes path operations descriptions from vocabulary annotations

irvinesunday:
Closes https://github.com/microsoft/OpenAPI.NET.OData/issues/103
Closes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/529

- Retrieves operation descriptions from vocabulary annotations of `IEdmVocabularyAnnotatable` elements.
- Tests have been updated to validate this feature.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/EntityGetOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/EntityGetOperationHandler.cs
@@ -30,14 +30,17 @@ namespace Microsoft.OpenApi.OData.Operation
             // Summary
             operation.Summary = "Get entity from " + EntitySet.Name + " by key";
 
+            IEdmEntityType entityType = EntitySet.EntityType();
+
+            // Description
+            operation.Description = Context.Model.GetDescriptionAnnotation(entityType);
+
             // OperationId
             if (Context.Settings.EnableOperationId)
             {
-                string typeName = EntitySet.EntityType().Name;
+                string typeName = entityType.Name;
                 operation.OperationId = EntitySet.Name + "." + typeName + ".Get" + Utils.UpperFirstChar(typeName);
             }
-
-            base.SetBasicInfo(operation);
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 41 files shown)


src/Microsoft.OpenApi.OData.Reader/Operation/EntityGetOperationHandler.cs
```diff
@@ -30,14 +30,17 @@ protected override void SetBasicInfo(OpenApiOperation operation)
             // Summary
             operation.Summary = "Get entity from " + EntitySet.Name + " by key";
 
+            IEdmEntityType entityType = EntitySet.EntityType();
+
```
xuzhg, line 34: There's a line to set the description in the base class.

I don't know why do you want to rewrite it using the entity type. 

Maybe you need to change that logic:

1) If we have description for the entity set, use it
2) otherwise, if we have description for entity type, use it
3) otherwise, leave it empty or use a default string.

xuzhg (changes requested):

src/Microsoft.OpenApi.OData.Reader/Operation/EntityGetOperationHandler.cs
irvinesunday: Are we not expecting a scenario where we have descriptions for both entity set and entity? In such a scenario, we would end up setting entity set descriptions for entities, even though they have their own annotated descriptions.

xuzhg: You mean for "single entity request, the annotation should from entity type?".

irvinesunday: Yes.

xuzhg: Typically, it should comes from entity set, not from entity type.
However, it seems we can't distinguish it from entity set or single entity.
So, I am ok for your changes.

xuzhg (approved):

Issue #107, opened 2021-07-05
Explicitly set parameters' explode property to false


irvinesunday:
When `style: form`, OData renders the query parameters: `orderby`, `select` and `expand` using `explode: false`, e.g., `$select=id,displayName,givenName`. We need to explicitly set this because in `Microsoft.OpenApi 1.3.0-preview` and above this is expected, otherwise `explode` defaults to `false`.

### Assemblies affected

`v1.0.8` and below

### Steps to reproduce

Bump up the `Microsoft.OpenApi` package to `v1.3.0-preview` and use the GUI tool to generate the OpenAPI doc. from the CSDL: https://graph.microsoft.com/v1.0/$metadata.
### Expected result

In `orderby`, `select` and `expand` path parameters we expect to see the property `explode: false`

### Actual result

The `explode` property of the above mentioned query params are missing. This is because they default to `true` and the OpenAPI writers exclude them from the generated doc. 

Related to: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/618

PR #108, opened 2021-07-05
Set parameters' explode property to false

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/107
Fixes microsoftgraph/microsoft-graph-devx-api#618

This PR:
- Explicitly sets the `Explode` properties to `false` for `$orderby`, `$select` and `$expand` OData query parameters, for `style: form`.
- Adds tests to validate the creation of the above query parameters, and in extension, validate that `explode: false` is always set.


xuzhg (approved):

xuzhg:
looks good to me. :shipit:

peombwa (approved):

Issue #114, opened 2021-09-14
Non-contained nav. props. should not have action/function paths


irvinesunday:
Non-contained navigation properties should not have action/function operations paths generated. This will reduce a number of redundant paths.

### Assemblies affected

`1.0.8`

### Steps to reproduce

Generate the OpenAPI doc. from CSDL with actions/functions bound to a type defining a navigation property that is con-contained.
### Expected result

No actions/functions path generated for this navigation property path.

### Actual result

actions/functions paths generated for this navigation property path.

PR #115, opened 2021-09-14
Skips generation of action/function paths for non-contained nav. props.

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/114

This PR: 
- Updates the `ODatPathProvider.cs` class to allow skipping of generation of action/function paths for non-contained navigation properties.
- Updates the test class with test methods to validate the above.


xuzhg (approved):
:shipit:

Issue #126, opened 2021-11-17
CSDL to OpenAPI Translation of Responses with Primitive Values


irvinesunday:
Refer to this issue: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/726

### Assemblies affected

`v1.0.9`

PR #127, opened 2021-11-17
Fixes response objects for operations that return single primitive type values

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/126

This PR:

- Fixes the schema of response objects of operations that return a single primitive type
- Refactors common code that creates `OpenApiResponses` for `IEdmOperation`s and `IEdmOperationImport`s into a common function that can be reused by both.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
@@ -4,6 +4,7 @@
 // ------------------------------------------------------------
 
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.OData.Edm;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.OData.Common;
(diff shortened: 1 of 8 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
```diff
@@ -72,35 +73,79 @@ public static OpenApiResponse GetResponse(this string statusCode)
...
+                                "value", context.CreateEdmTypeSchema(operation.ReturnType)
+                            }
+                        }
+                    };
+                }
+                else
+                {
+                    // A property or operation response that is of a primitive type is represented as an object with a single name/value pair,
+                    // whose name is value and whose value is a primitive value.
+                    if (operation.ReturnType.IsPrimitive())
```
peombwa, line 129: nit: This can be simplified with an `else if` since we are also checking the `operation.ReturnType` in the `if` above.

peombwa (approved):

PR #143, opened 2021-11-23
bugfix/missing quotes templates

### LINKED ISSUE #129: Count path segments are missing after translation
[OData supports counts segements](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingtheCountofaCollection) by conventions. e.g. GET /users/$count

All these segments are missing from the resulting OpenAPI description.

### LINKED ISSUE #140: URL templates missing quotes around enum types
According to the spec, it seems that we need quotes around the enum type name but not the namespace:
http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_PrimitiveLiterals
> `ColorEnumValue eq Sales.Pattern'Yellow',`
> `ColorEnumValue eq 'Yellow',`

These quotes are missing in the translated OpenAPI doc.

see https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1156 for more details.

baywet:
- fixes #140 a bug where quotes would be missing in url templates

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
@@ -7,29 +7,29 @@ using Microsoft.OData.Edm;
 
 namespace Microsoft.OpenApi.OData.Edm
 {
-	/// <summary>
+    /// <summary>
     /// Extension methods for <see cref="IEdmType"/>
     /// </summary>
     public static class EdmTypeExtensions
-	{
-		/// <summary>
-		/// Determines wether a path parameter should be wrapped in quotes based on the type of the parameter.
-		/// </summary>
-		/// <param name="edmType">The type of the parameter.</param>
-		public static bool ShouldPathParameterBeQuoted(this IEdmType edmType)
-		{
-			if (edmType == null)
-			{
-				return false;
-			}
+    {
+        /// <summary>
+        /// Determines wether a path parameter should be wrapped in quotes based on the type of the parameter.
+        /// </summary>
+        /// <param name="edmType">The type of the parameter.</param>
+        public static bool ShouldPathParameterBeQuoted(this IEdmType edmType)
+        {
+            if (edmType == null)
+            {
+                return false;
+            }
 
-			return edmType.TypeKind switch
-			{
-				EdmTypeKind.Enum => true,
-				EdmTypeKind.Primitive when edmType.IsString() || edmType.IsTemporal() => true,
-				_ => false,
-			};
-		}
-	}
+            return edmType.TypeKind switch
+            {
+                EdmTypeKind.Enum => true,
+                EdmTypeKind.Primitive when edmType.IsString() || edmType.IsTemporal() => true,
+                _ => false,
+            };
+        }
+    }
 
 }
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
@@ -1,0 +1,35 @@
+// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// ------------------------------------------------------------
+
+using Microsoft.OData.Edm;
+
+namespace Microsoft.OpenApi.OData.Edm
+{
+	/// <summary>
+    /// Extension methods for <see cref="IEdmType"/>
+    /// </summary>
+    public static class EdmTypeExtensions
+	{
+		/// <summary>
+		/// Determines wether a path parameter should be wrapped in quotes based on the type of the parameter.
+		/// </summary>
+		/// <param name="edmType">The type of the parameter.</param>
+		public static bool ShouldPathParameterBeQuoted(this IEdmType edmType)
+		{
+			if (edmType == null)
+			{
+				return false;
+			}
+
+			return edmType.TypeKind switch
+			{
+				EdmTypeKind.Enum => true,
+				EdmTypeKind.Primitive when edmType.IsString() || edmType.IsTemporal() => true,
+				_ => false,
+			};
+		}
+	}
+
+}
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
@@ -16,9 +16,11 @@ namespace Microsoft.OpenApi.OData.Edm
         /// Determines wether a path parameter should be wrapped in quotes based on the type of the parameter.
         /// </summary>
         /// <param name="edmType">The type of the parameter.</param>
-        public static bool ShouldPathParameterBeQuoted(this IEdmType edmType)
+        /// <param name="settings">The conversion settings.</param>
+        /// <returns>True if the parameter should be wrapped in quotes, false otherwise.</returns>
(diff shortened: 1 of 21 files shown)


test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.yaml
```diff
@@ -390,7 +390,7 @@ paths:
         default:
           $ref: '#/components/responses/error'
       x-ms-docs-operation-type: operation
-  '/GetNearestAirport(lat={lat},lon={lon})':
+  '/GetNearestAirport(lat=''{lat}'',lon=''{lon}'')':
```
zengin, line 393: same comment but for double.

zengin (changes requested):

test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataOperationSegmentTests.cs
```diff
@@ -74,10 +74,10 @@ public void GetPathItemNameReturnsCorrectActionLiteral(bool unqualifiedCall, boo
         }
 
         [Theory]
-        [InlineData(true, true, "MyFunction(param={param})")]
-        [InlineData(true, false, "MyFunction(entity=@entity,param={param})")]
-        [InlineData(false, true, "NS.MyFunction(param={param})")]
-        [InlineData(false, false, "NS.MyFunction(entity=@entity,param={param})")]
+        [InlineData(true, true, "MyFunction(param='{param}')")]
```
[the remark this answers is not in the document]
zengin: http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_PrimitiveLiterals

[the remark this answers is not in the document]
xuzhg: single quote is used for string primitive type value in the request path. 
for the template, I don't understand why should we have a single quote around.
It's just a template and anything within the {} is the parameter name used to retrieve the value from request.

xuzhg (changes requested):

xuzhg:
`MyFunction(param={param})` can match

- ~ MyFunction(param='abc')  // if the type of param is string
- ~ MyFunction(param=true) // if the type of param is boolean
- ~ MyFunction(param=8) // if the type of param is int32
- ...
- 

MyFunction(param='{param}') can matches ?

- ~ MyFunction(param=''abc'')  // if the type of param is string
- ~ MyFunction(param='true') // if the type of param is boolean
- ~ MyFunction(param='8') // if the type of param is int32

test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataOperationSegmentTests.cs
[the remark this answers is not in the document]
baywet: thank you both for catching my replacement mistake! I reviewed my copy and just pushed a version which I believe will be more accurate, let me know what you think.

zengin (approved):

src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+	/// <summary>
+    /// Extension methods for <see cref="IEdmType"/>
+    /// </summary>
+    public static class EdmTypeExtensions
+	{
+		/// <summary>
+		/// Determines wether a path parameter should be wrapped in quotes based on the type of the parameter.
+		/// </summary>
+		/// <param name="edmType">The type of the parameter.</param>
```
xuzhg, line 18: format

src/Microsoft.OpenApi.OData.Reader/Edm/ODataKeySegment.cs
```diff
@@ -77,13 +77,7 @@ public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<
                 }
                 else
                 {
-                    IList<string> keyStrings = new List<string>();
```
xuzhg, line 80: Again, I have a different opinion for the single quote.

Single quote is part of the string, enum value, it goes with value. But not for the template.

When Graph SDK generate the URL (the real URL), should generate the single quote for string type. 
For example:

'8' for string type, but 8 for int type. 

So, the changes in this PR looks "invalid" to me. I'd like @mikepizzo's input.

baywet: I disagree. This quotes behavior comes from odata conventions. The open api spec does not mandate any quote for url parameters, regardless of the types.
The point of this library is to encapsulate the odata conventions away.
This is why I believe the quotes should be added during the conversion process.

src/Microsoft.OpenApi.OData.Reader/Edm/EdmTypeExtensions.cs
baywet: Sorry, what's the issue with the format here?

xuzhg: alignment, thanks.

baywet: are you referring to the use of tabs vs spaces? still trying to understand what you think needs to be changed here.

xuzhg: Yes, replace tab using 4 spaces.

baywet: done!

darrelmiller:
The OpenAPI that we need to generate should look like this,

```yaml
openapi: 3.0.0
info:
  title: example of string path parameter that must be quoted
  version: 1.0.0
paths:
  get:
    /MyFunction(param='{funcParam}'):
      parameters:
        - name: funcParam
          in: path
          required: true
          schema:
             type: string
      responses: 
        200:
          description: OK
```

JSON Schema has no notion of a quoted string, so it would be difficult to describe a parameter value that needs to be surrounded by quotes before inserting into the URL.  The fact that quotes, single vs double are required are an OData requirement and therefore should be encoded in URL, just like all other OData conventions.

You can see in the simple string expansion examples in the URI Template specification that quotes are not inserted for string values.

```code
     Example Template     Expansion

       {var}              value
       {hello}            Hello%20World%21
       {half}             50%25
       O{empty}X          OX
       O{undef}X          OX
       {x,y}              1024,768
       {x,hello,y}        1024,Hello%20World%21,768
       ?{x,empty}         ?1024,
       ?{x,undef}         ?1024
       ?{undef,y}         ?768
       {var:3}            val
       {var:30}           value
       {list}             red,green,blue
       {list*}            red,green,blue
       {keys}             semi,%3B,dot,.,comma,%2C
       {keys*}            semi=%3B,dot=.,comma=%2C
```
https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.2

baywet:
collecting notes from the meeting we've just had:

- clients rely on this conversion library to encapsulate away OData conventions and therefore have no knowledge of encoding requirements (quotes)
- the information (quotes) cannot be carried through JSON schema/OpenAPI structures
- this should work in 99.999% of the times, except maybe for Edm.Primitive types which are not allowed in url templates
- the tradeoff of this solution is making the template a little more static in nature
- an alternative could be considered to use the format property, and use an open formatter (odatavalue, odatastring...) but this brings other concerns like collisions with numbers which already use the format property.

Overall the group agreed to proceed with this change.

@xuzhg for final review and merge.

xuzhg:
:shipit:

Issue #142, opened 2021-11-23
Configure repository for static analysis + code linting


baywet:
A few comments in #141 started a conversation around static code analysis and code linting.
Ideally we want to enforce using the latest and most performant code patterns, catch any code syntax bug and enforce consistent writing style.
This is hard to enforce mechanically as no tooling is configured on the repository today, making catching these "errors" tedious.

## On static code analysis

Requirements:

- should not require the full version of Visual Studio or any specific editor
- should run with CI so we can enforce the check on pull requests

Suggested solutions:

- https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers/
- https://www.nuget.org/packages/StyleCop.Analyzers/1.2.0-beta.354
- https://www.nuget.org/packages/SonarAnalyzer.CSharp/ or https://sonarcloud.io
- https://www.nuget.org/packages/Microsoft.CodeAnalysis.Analyzer.Testing/

Personal opinion:

I've personally been very happy with sonarcloud (the service) through the years which comes for free for any open source project. Readers [can explore Kiota on sonarcloud](https://sonarcloud.io/project/issues?id=microsoft_kiota) to get an idea of the kind of issues it catches.

And nothing prevents us from mix and matching as long as rules don't conflict.

## On code linting

Requirements:

- should not require the full version of Visual Studio or any specific editor
- should run with CI so we can enforce the check on pull requests
- should provide live feedback in the editor
- should provide a way to "mass-resolve" infringements

Suggested solutions:

- https://www.nuget.org/packages/StyleCop.Analyzers/1.2.0-beta.354 with https://marketplace.visualstudio.com/items?itemName=ChrisDahlberg.StyleCop in VS and the CLI in vs code
- Resharper with https://marketplace.visualstudio.com/items?itemName=fakesharper.fakesharper for VS code
- editor config https://asp.net-hacker.rocks/2020/01/13/editorconfig.html (see the [comprehensive rules of raptor](https://github.com/microsoftgraph/msgraph-sdk-raptor/blob/dev/.editorconfig) )

More reading https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview

Personal opinion:

I've had a good experience with most of those with a preference for StyleCop and/or editor config.

PR #149, opened 2021-11-24
type cast segments for derived types

### LINKED ISSUE #124: Missing OData cast path items
`/groups/{id}/members` is a collection of directory objects.
The service supports downcasting the result by using the following endpoint `/groups/{id}/members/microsoft.graph.user` which not only filters down the collection to objects of types users, but also casts the items in the returned collection.

All cast endpoints are missing from the resulting description.

baywet:
fixes #124

## Summary

This pull request adds two new settings `EnableODataTypeCast` (default true) and `RequireDerivedTypesConstraintForODataTypeCastSegments` (default true) and adds type cast segments for singletons, entity sets, entity set items (keys), navigation properties bound to a single item, navigation properties items (key) of types which have derived types.
It also adds child type operations and child type navigation properties under the singleton, entity set item, single nav prop, and nav prop item cast segments.
It also adds count segments under cast segment that are under collection navigation properties and entity sets.
With examples: 

- ~/singleton/ns.typecast + (derived type nav props, derived type operations)
- ~/.../singlenavproperty/ns.type + (derived type nav props, derived type operations)
- ~/.../multiplenavproperty/ns.type + (derived type collection operations, count)
- ~/.../multiplenavproperty/{id}/ns.type + (derived type nav props, derived type operations)
- ~/entityset/ns.type + (derived type collection operations, count)
- ~/entityset/{id}/ns.type + (derived type nav props, derived type operations)

It will read the derived type constraints annotations for example

```xml
<Annotation Term="Org.OData.Validation.V1.DerivedTypeConstraint">
          <Collection>
            <String>microsoft.graph.user</String>
            <String>microsoft.graph.group</String>
            <String>microsoft.graph.application</String>
            <String>microsoft.graph.servicePrincipal</String>
            <String>microsoft.graph.device</String>
            <String>microsoft.graph.orgContact</String>
          </Collection>
        </Annotation>
```

The logic of whether the segments will be added or not can be exhaustively described with

| Enable | Require | Annotation | Found in annotation | Result |
| ------ | ------- | ---------- | ------------------- | ------ |
| true | true | true | true | Added |
| true | true | true | false | Noop |
| true | true | false | true | Impossible |
| true | true | false | false | Noop |
| true | false | true | true | Added |
| true | false | true | false | Noop |
| true | false | false | true | Impossible |
| true | false | false | false | Added |
| false | true | true | true | Noop |
| false | true | true | false | Noop |
| false | true | false | true | Noop |
| false | true | false | false | Noop |
| false | false | true | true | Noop |
| false | false | true | false | Noop |
| false | false | false | true | Noop |
| false | false | false | false | Noop |

Or segments will only be added when the option is enabled and either:
- annotation is required, present and contains the derived type name
- annotation is not required, and (either annotation is present with the type name or absent)


darrelmiller:
Will this generate paths for navigation properties that only exist on derived classes?  

image

Ideally we want to be able to do something like,
```csharp
   graphClient.Teams[id].Members.AsAadUser.User.GetAsync()
```

baywet:
there's still work to get there but that the goal yes.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPath.cs
```diff
@@ -315,28 +317,31 @@ private ODataPathKind CalcPathType()
...
             else if (Segments.Count == 2 && Segments.Last().Kind == ODataSegmentKind.Key)
             {
                 return ODataPathKind.Entity;
             }
 
             return ODataPathKind.Unknown;
         }
+
+        /// <summary>
+        /// Profides a suffix for the operation id based on the operation path.
```
xuzhg, line 340: typo?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPath.cs
```diff
@@ -315,28 +317,31 @@ private ODataPathKind CalcPathType()
...
-                        return ODataPathKind.EntitySet;
-                    }
+                    return ODataPathKind.Singleton;
                 }
-            }
+                else
+                {
+                    return ODataPathKind.EntitySet;
+                }
+			}
```
xuzhg, line 330: replace tab using 4 white spaces.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathKind.cs
```diff
@@ -60,9 +60,14 @@ public enum ODataPathKind
         /// </summary>
         DollarCount,
 
+        /// <summary>
+        /// Represents a type cast path, for example: ~/groups/{id}/members/microsoft.graph.user
+        /// </summary>
+		TypeCast,
```
xuzhg, line 66: alignment (replace tab using 4 white spaces)

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -393,6 +413,60 @@ private void CreateCountPath(ODataPath currentPath, OpenApiConvertSettings conve
...
+        /// Create OData type cast paths.
+        /// </summary>
+        /// <param name="currentPath">The current OData path.</param>
+        /// <param name="convertSettings">The settings for the current conversion.</param>
+        /// <param name="structuredType">The type that is being inherited from to which this method will add downcast path segments.</param>
+        /// <param name="annotable">The annotable navigation source to read cast annotations from.</param>
+        /// <param name="targetsMany">Whether the annotable navigation source targets many entities.</param>
+        private void CreateTypeCastPaths(ODataPath currentPath, OpenApiConvertSettings convertSettings, IEdmStructuredType structuredType, IEdmVocabularyAnnotatable annotable, bool targetsMany)
+        {
+            if(currentPath == null) throw new ArgumentNullException(nameof(currentPath));
```
xuzhg, line 426:
```C#
if (...)
{
     throw Error.ArgumentNull(...)
}
```

for readability

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -393,6 +413,60 @@ private void CreateCountPath(ODataPath currentPath, OpenApiConvertSettings conve
...
+            if(annotable == null) throw new ArgumentNullException(nameof(annotable));
+            if(!convertSettings.EnableODataTypeCast) return;
+
+            var annotedTypeNames = GetDerivedTypeConstaintTypeNames(annotable);
+            
+            if(!annotedTypeNames.Any() && convertSettings.RequireDerivedTypesConstraintForODataTypeCastSegments) return; // we don't want to generate any downcast path item if there is no type cast annotation.
+
+            var annotedTypeNamesSet = new HashSet<string>(annotedTypeNames, StringComparer.OrdinalIgnoreCase);
+
+			bool filter(IEdmStructuredType x) =>
```
xuzhg, line 438: aligment again?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -393,6 +413,60 @@ private void CreateCountPath(ODataPath currentPath, OpenApiConvertSettings conve
...
+        {
+            if(currentPath == null) throw new ArgumentNullException(nameof(currentPath));
+            if(convertSettings == null) throw new ArgumentNullException(nameof(convertSettings));
+            if(structuredType == null) throw new ArgumentNullException(nameof(structuredType));
+            if(annotable == null) throw new ArgumentNullException(nameof(annotable));
+            if(!convertSettings.EnableODataTypeCast) return;
+
+            var annotedTypeNames = GetDerivedTypeConstaintTypeNames(annotable);
+            
+            if(!annotedTypeNames.Any() && convertSettings.RequireDerivedTypesConstraintForODataTypeCastSegments) return; // we don't want to generate any downcast path item if there is no type cast annotation.
```
xuzhg, line 434: don't need to check the constaint items? 
I mean to check whether the constraint can apply to a certain type cast?

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -188,6 +188,16 @@ public string PathPrefix
         /// </summary>
         public bool EnableDollarCountPath { get; set; } = true;
 
+        /// <summary>
+        /// Gets/sets a value indicating whether or not to include the OData type cast segments.
+        /// </summary>
+        public bool EnableODataTypeCast { get; set; } = true;
```
xuzhg, line 194: It's better to disable it by default.

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -188,6 +188,16 @@ public string PathPrefix
...
+
+        /// <summary>
+        /// Gets/sets a value indicating whether or not to require a derived types constraint to include the OData type cast segments.
+        /// </summary>
+        public bool RequireDerivedTypesConstraintForODataTypeCastSegments { get; set; } = true;
```
xuzhg, line 199: It's better to disable it by default.

test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/ODataTypeCastGetOperationHandlerTests.cs
```diff
@@ -0,0 +1,352 @@
...
+using Microsoft.OpenApi.Extensions;
+using Microsoft.OpenApi.OData.Edm;
+using Microsoft.OpenApi.OData.PathItem.Tests;
+using Microsoft.OpenApi.OData.Tests;
+using Xunit;
+
+namespace Microsoft.OpenApi.OData.Operation.Tests;
+public class ODataTypeCastGetOperationHandlerTests
+{
+  private readonly ODataTypeCastGetOperationHandler _operationHandler = new ();
```
xuzhg, line 17: format, make it readable.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
```diff
@@ -100,6 +102,17 @@ public string GetPathItemName(OpenApiConvertSettings settings)
         {
             return GetPathItemName(settings, new HashSet<string>());
         }
+        /// <summary>
+        /// Profides a suffix for the operation id based on the operation path.
+        /// </summary>
+        /// <param name="path">Path to use to deduplicate.</param>
+        /// <param name="settings">The settings.</param>
+        ///<returns>The suffix.</returns>
+        public string GetPathHash(OpenApiConvertSettings settings, ODataPath path = default)
```
xuzhg, line 111: no added into the public api?

xuzhg:
Excuse me. But, Coding style, clean coding, making code readable... are my style. Thanks.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPath.cs
baywet:
```suggestion
        /// Provides a suffix for the operation id based on the operation path.
```

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
baywet: because the require one is true, this will be noop as long as no annotations are present in the CDSL

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
baywet: this true by default means we won't add any path segment if no annotations are present

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
baywet:
```suggestion
            if(currentPath == null) throw Error.ArgumentNull(nameof(currentPath));
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPath.cs
baywet: done

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathKind.cs
baywet: done

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
baywet: done

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
baywet: the the consumer requires the annotation before adding any type cast segment, and there's no annotation present, there's no point continuing further

src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
baywet: it was, unless I missed something? https://github.com/microsoft/OpenAPI.NET.OData/pull/149/files#diff-2b4bfbc25273377536cf444393058ae0e6da90188a25e7cfb19caff894997678R70

baywet:
> Excuse me. But, Coding style, clean coding, making code readable... are my style. Thanks.

The "Excuse me" part of this passive aggressive comment does not make it more acceptable. IDEs and people are going to have different configurations and preferences. If you care about tab, spaces, curlies and other details like that, please take the time to push a configuration that's enforced through tooling as described in #142 as as we've discussed offline.
Until some systematic linting is in place, arguing about those things and manually fixing those is a waste of time.

Now, can we focus on reviewing the code rather than the whites-pacing around the code please?

xuzhg:
:shipit:

Issue #151, opened 2021-12-08
[Feature] Uses Referenceable annotation to control generation of $ref paths


irvinesunday:
We want to introduce a `Referenceable` `Org.OData.Capabilities.V1.NavigationRestrictions` annotation for controlling the generation of `$ref` paths. This shall be used instead of relying on `ContainsTarget=false` or the absence of this attribute in a navigation property as a condition for generating `$ref` paths.

baywet:
related #123

PR #152, opened 2021-12-08
Uses Referenceable annotation to control generation of $ref paths

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/151
Fixes https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/754

This PR proposes:
- Introducing and using a `Referenceable` annotation to control the generation of `$ref` paths. This helps in the separation of using `ContainsTarget=false` or the absence of this attribute in a navigation property as a condition for generating `$ref` paths. 
- Updating the relevant tests and integration test files to validate the above.


test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Graph.Beta.OData.xml
```diff
@@ -2502,7 +2502,7 @@
                 <NavigationProperty Name="items" Type="Collection(microsoft.graph.driveItem)" ContainsTarget="true" />
                 <NavigationProperty Name="list" Type="microsoft.graph.list" ContainsTarget="true" />
                 <NavigationProperty Name="root" Type="microsoft.graph.driveItem" ContainsTarget="true" />
-                <NavigationProperty Name="special" Type="Collection(microsoft.graph.driveItem)" ContainsTarget="true" />
+                <NavigationProperty Name="special" Type="Collection(microsoft.graph.driveItem)" ContainsTarget="true" sdk:Explode="false"/>
```
xuzhg, line 2505: Is it related to OData? 
I'd suggest creating a new branch for "Graph SDK", and doing a separate release for it.

@mikepizzo any input?

baywet: do we have a standard annotation/attribute we could be using instead? I know this is something we've discussed about at lengths in the API council to split the information of ContainsTarget into separate attributes to clarify some scenarios like this one.
Mike's input definitively needed. Also probably @darrelmiller can help here.

darrelmiller: @xuzhg This isn't specifically an OData description issue. This is a problem of the fact that if CSDL reuses of a large type in many places within a CSDL description can end up generating a large number of unnecessary paths.  Currently this library uses the heuristic that a top level entityset would prevent the exploding of paths for navigation properties of the same type.  That isn't always true in OData/CSDL but it was good enough for most cases.  We have now run in a case where that isn't good enough.

In Graph's case DriveItem has many nav properties and many functions bound to it.  It also appears in many places in Graph.  OneDrive/SharePoint implement many of the paths for convenience, but we don't want to generate all of the them in OpenAPI for every instance of DriveItem, because all DriveItems can be accessed from the one main path /drives/{driveId}/items/{driveItemId}.

This XML based attribute is a hint to help people constrain the projection of OpenAPI paths from OData.  In is necessary because CSDL is convention based and OpenAPI isn't.  This is not a Graph specific thing.  This is going to be a useful for anyone with a non-trivial CSDL.

chrisspre: a) is there a specific reason to use the XML based attribute instead of a custom annotation 
```xml

    <NavigationProperty Name="special" Type="Collection(microsoft.graph.driveItem)" ContainsTarget="true">
          <Annotation Term="microsoft.graph.Explode" Bool="true" />      
    </NavigationProperty>
    
b) There is the Cabability called ExpandRestrictions in hte standard cabailities vocabulary that list the non-expandable navigation properties. But the definition is, that query options to expand that property are not accepted. I am not 100% sure if that is exactly the same as what you seem to need.

Comment (darrelmiller) [reply test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Graph.Beta.OData.xml:2505]: The primary reason that I suggested custom annotations is because, the annotation isn't describing anything about the API itself, but is a directive to tooling about how to process the API description.  The annotation tells process that determines what OpenAPI paths should be generated to not create paths for any navigation properties.  This is unrelated to the ability to expand navigation properties.

Comment (baywet) [reply test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Graph.Beta.OData.xml:2505]: I here is some additional information I was able to get from @mikepizzo internally

> [navigationRestriction](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#NavigationRestrictionsType) that can be used to say whether the service supports navigating the relationship that could be used to tell the tool whether or not to generate the path for the navigation property. However,
> 1. If the relationship is containment, then we should always assume it's navigable since that is the canonical URL for the item, and
> 1. Even if the relationship is navigable, we might not want to advertise it in OpenAPI simply because doing so may make the OpenAPI definition untenably large.
>
> To try and mitigate the size of the OpenAPI, in our initial design we said we would expand all paths through every canonical url (i.e., entity set + containment navigation paths, recursively) plus one level. The rationale for this was:
> 1. We should at least include the containment path, for the reason stated in 1 above
> 1. We should go one navigation past the entity set/containment path because, if you can't navigate the relationship from its canonical URL then it's not much of a relationship)
> 1. We didn't need to go further because, for any other path, there is already a (more direct) path to get there (i.e, the one above). So additional paths may be valid, but were of increasingly diminishing value so not worth bloating the OpenAPI definition.

So here is my suggestion: let's switch to that standard annotation instead and wrap up this PR, thoughts?

Comment (irvinesunday) [reply test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Graph.Beta.OData.xml:2505]: @baywet what standard annotation are you suggesting be used here as a flag to turn off expanding containment navigation properties?

Comment (baywet) [reply test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Graph.Beta.OData.xml:2505]: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#NavigationRestrictionsType

Comment (mikepizzo) [reply test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Graph.Beta.OData.xml:2505]: Is the issue here that the ContainsTarget attribute is being misused in MS Graph?

If ContainsTarget is in fact true, I would expect we should always include the path, as it indicates the canonical URL.

In the DriveItems case, is /drives/{driveId}/items/{driveItemId} the canonical ID?

Are there other examples where we don't want to expand a containment path?

File: src/Microsoft.OpenApi.OData.Reader/Edm/EdmModelExtensions.cs [hunk 2]
```diff
@@ -152,6 +152,7 @@ public static IEnumerable<IEdmComplexType> FindAllBaseTypes(this IEdmComplexType
         /// <param name="baseType">Type of the base type.</param>
         /// <param name="subtype">Type of the sub type.</param>
         /// <returns>true, if the <paramref name="baseType"/> is assignable to <paramref name="subtype"/>. Otherwise returns false.</returns>
+        [Obsolete]
```
baywet, line 155: why is this method tagged as obsolete? since it's part of the public API, we should at least provide a rationale

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Edm/EdmModelExtensions.cs
irvinesunday: This shouldn't have been here, this was made obsolete in this other [PR 160](https://github.com/microsoft/OpenAPI.NET.OData/pull/160), from this [suggestion](https://www.youtube.com/watch?v=-0uPUgdTbnw). I have corrected it.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -426,7 +426,7 @@ private void RetrieveNavigationPropertyPaths(IEdmNavigationProperty navigationPr
             currentPath.Pop();
         }
 
-        private bool ShouldExpandNavigationProperty(IEdmNavigationProperty navigationProperty, ODataPath currentPath)
+        private bool ShouldExpandNavigationProperty(IEdmNavigationProperty navigationProperty, ODataPath currentPath, OpenApiConvertSettings convertSettings)
```
baywet, line 429: the added parameter is not used? remove it

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -366,9 +366,9 @@ private void RetrieveNavigationPropertyPaths(IEdmNavigationProperty navigationPr
                 // ~/entityset/{key}/single-valued-Nav/subtype
                 CreateTypeCastPaths(currentPath, convertSettings, navigationProperty.DeclaringType, navigationProperty, targetsMany);
 
-                if (!navigationProperty.ContainsTarget)
+                if (navigation?.Referenceable == true)
```
baywet, line 369: this translates to requiring explicit annotations.
i.e. if no annotation value or if an annotation value is present and false, don't add the path items,
if if an annotation value is present and true, add the path items.
Which will require workloads to explicitly tell us the ref endpoints are supported.
I personally don't mind but wanted to check first with @mikepizzo and @darrelmiller whether this was the desired behavior?

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
darrelmiller: I would prefer us to be opt-in.  I would rather add the paths that we know are implemented than remove those that are not.  It's not the original OData philosophy but it is more natural for OpenAPI generation.

irvinesunday: @darrelmiller what does your suggestion mean in terms of implementation? Do we check explicitly for `Referenceable=true` annotations before creating `$ref` paths and nullability or `Referenceable=false` means no `$ref` paths to be created?

darrelmiller: $ref paths should only be created if Referenceable is present and true.  I don't know why nullability is relevant.

baywet: thanks for clarifying the intent here, no changes required, resolving

baywet (approved):

PR #155, opened 2021-12-13
adds support for deprecation

### LINKED ISSUE #113: Set deprecated property for operations that have been deprecated

baywet:
fixes #113

This PR sets `deprecated` to `true` on operations that depend on annotables with an annotation of term `Org.OData.Core.V1.Revisions` and of kind `Org.OData.Core.V1.RevisionKind/Deprecated` (see the linked issue for more details).

It will also add and extension to the operation to convey the additional information from the annotation like

```yaml
/People/$count:
    get:
      summary: Get the number of the resource
      operationId: Get.Count.People
      responses:
        '200':
          description: The count of the resource
          content:
            text/plain:
              schema:
                $ref: '#/components/schemas/ODataCountResponse'
        default:
      deprecated: true
      x-ms-deprecation:
        removalDate: '2023-03-15T00:00:00.0000000'
        date: '2021-08-24T00:00:00.0000000'
        version: 2021-05/people
        description: The People API is deprecated and will stop returning data on March 2023. Please use the new newPeople API.
```

The flag + extension will be added to any operation if any path segment in the path is annotated including:
- parent entity sets
- parent singleton
- parent navigation properties
- the type or base type of the nav prop/singleton/entity set

> Note: depending on merge order, #149 will need to be updated to return the target type in the annotables for the segment

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataDollarCountSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataDollarCountSegment.cs
@@ -25,6 +25,7 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <inheritdoc />
         public override string Identifier => "$count";
 
+        /// <inheritdoc />
 		public override IEnumerable<IEdmVocabularyAnnotatable> GetAnnotables()
 		{
 			return Enumerable.Empty<IEdmVocabularyAnnotatable>();
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataDollarCountSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataDollarCountSegment.cs
@@ -4,6 +4,8 @@
 // ------------------------------------------------------------
 
 using System.Collections.Generic;
+using System.Linq;
+using Microsoft.OData.Edm.Vocabularies;
 
 namespace Microsoft.OpenApi.OData.Edm
 {
@@ -23,7 +25,12 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <inheritdoc />
         public override string Identifier => "$count";
 
-        /// <inheritdoc />
-        public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<string> parameters) => "$count";
+		public override IEnumerable<IEdmVocabularyAnnotatable> GetAnnotables()
+		{
+			return Enumerable.Empty<IEdmVocabularyAnnotatable>();
+		}
+
+		/// <inheritdoc />
+		public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<string> parameters) => "$count";
     }
 }
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
@@ -7,8 +7,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OData.Edm;
-using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.OData.Common;
+using Microsoft.OpenApi.OData.OpenApiExtensions;
 
 namespace Microsoft.OpenApi.OData.Edm
 {
@@ -121,5 +121,9 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <param name="parameters">The existing parameters.</param>
         /// <returns>The path item name.</returns>
         public abstract string GetPathItemName(OpenApiConvertSettings settings, HashSet<string> parameters);
+        /// <summary>
+        /// Provides any deprecation information for the segment.
+        /// </summary>
+        public OpenApiDeprecationExtension Depreciation { get; set; }
     }
 }
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
@@ -124,6 +124,6 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <summary>
         /// Provides any deprecation information for the segment.
         /// </summary>
-        public OpenApiDeprecationExtension Depreciation { get; set; }
+        public OpenApiDeprecationExtension Deprecation { get; set; }
     }
 }
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OData.Edm;
 using Microsoft.OpenApi.OData.Common;
+using Microsoft.OData.Edm.Vocabularies;
 using Microsoft.OpenApi.OData.OpenApiExtensions;
 
 namespace Microsoft.OpenApi.OData.Edm
@@ -125,5 +126,9 @@ namespace Microsoft.OpenApi.OData.Edm
         /// Provides any deprecation information for the segment.
         /// </summary>
         public OpenApiDeprecationExtension Deprecation { get; set; }
+        /// <summary>
+        /// Returns the list of <see cref="IEdmVocabularyAnnotatable"/> this segment refers to.
+        /// </summary>
+        public abstract IEnumerable<IEdmVocabularyAnnotatable> GetAnnotables();
     }
 }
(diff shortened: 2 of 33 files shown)


test/Microsoft.OpenAPI.OData.Reader.Tests/OpenApiExtensions/OpenApiDeprecationExtensionTests.cs
```diff
@@ -0,0 +1,73 @@
+using System;
```
xuzhg, line 1: new files need the copyright header.

test/Microsoft.OpenAPI.OData.Reader.Tests/OpenApiExtensions/OpenApiDeprecationExtensionTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    [Fact]
+    public void ExtensionNameMatchesExpected()
+    {
+        // Arrange
+        OpenApiDeprecationExtension extension = new();
+
+        // Act
+        string name = extension.Name;
+        string expectedName = "x-ms-deprecation";
+        Assert.Equal(expectedName, name);
```
xuzhg, line 20: why duplicate the Assert.Equal(...)

test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/EdmFunctionOperationHandlerTests.cs
```diff
@@ -16,7 +16,58 @@ namespace Microsoft.OpenApi.OData.Operation.Tests
...
+          var operation = _operationHandler.CreateOperation(context, path);
+
+          // Assert
+          Assert.NotNull(operation);
+          Assert.True(operation.Deprecated);
+          Assert.NotEmpty(operation.Extensions);
+          var extension = operation.Extensions["x-ms-deprecation"];
+          Assert.NotNull(extension);
+        }
+        [Fact]
```
xuzhg, line 45: insert a new line after the test

test/Microsoft.OpenAPI.OData.Reader.Tests/Vocabulary/Capabilities/DeprecatedRevisionTypeTests.cs
```diff
@@ -0,0 +1,81 @@
+using System;
```
xuzhg, line 1: copyright header

test/Microsoft.OpenAPI.OData.Reader.Tests/Vocabulary/Capabilities/DeprecatedRevisionTypeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+using Microsoft.OData.Edm;
+using Microsoft.OData.Edm.Vocabularies;
+using Microsoft.OpenApi.OData.Vocabulary.Core;
+using Xunit;
+
+namespace Microsoft.OpenApi.OData.Reader.Vocabulary.Capabilities.Tests;
```
xuzhg, line 7: the namespace looks have different pattern comparing to the existing?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataDollarCountSegment.cs
```diff
@@ -24,6 +26,12 @@ public class ODataDollarCountSegment : ODataSegment
         public override string Identifier => "$count";
 
         /// <inheritdoc />
-        public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<string> parameters) => "$count";
+		public override IEnumerable<IEdmVocabularyAnnotatable> GetAnnotables()
```
xuzhg, line 29: alignment the codes, tab -> 4 whitespaces

xuzhg: It seems you can implement the default "GetAnnotables()" at the base class to return an empty Enumerable. Only override the method in the class that should be overridden?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
```diff
@@ -108,5 +109,13 @@ public string GetPathItemName(OpenApiConvertSettings settings)
         /// <param name="parameters">The existing parameters.</param>
         /// <returns>The path item name.</returns>
         public abstract string GetPathItemName(OpenApiConvertSettings settings, HashSet<string> parameters);
+        /// <summary>
```
xuzhg, line 112: add lines between blocks

src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
```diff
@@ -108,5 +109,13 @@ public string GetPathItemName(OpenApiConvertSettings settings)
...
+        /// Provides any deprecation information for the segment.
+        /// </summary>
+        public OpenApiDeprecationExtension Deprecation { get; set; }
```
xuzhg, line 115: The deprecation looks for the ODataPath, not a certain Segment in the Path, right?

src/Microsoft.OpenApi.OData.Reader/OpenApiExtensions/OpenApiDeprecationExtension.cs
```diff
@@ -0,0 +1,58 @@
+using System;
```
xuzhg, line 1: copyright header

test/Microsoft.OpenAPI.OData.Reader.Tests/OpenApiExtensions/OpenApiDeprecationExtensionTests.cs
baywet: solved

test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/EdmFunctionOperationHandlerTests.cs
baywet: solved

test/Microsoft.OpenAPI.OData.Reader.Tests/Vocabulary/Capabilities/DeprecatedRevisionTypeTests.cs
baywet: solved

src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
baywet: solved

src/Microsoft.OpenApi.OData.Reader/OpenApiExtensions/OpenApiDeprecationExtension.cs
baywet: solved

test/Microsoft.OpenAPI.OData.Reader.Tests/OpenApiExtensions/OpenApiDeprecationExtensionTests.cs
baywet: I got too enthusiastic with GitHub copilot, solved.

test/Microsoft.OpenAPI.OData.Reader.Tests/Vocabulary/Capabilities/DeprecatedRevisionTypeTests.cs
baywet: if it were up to me, I'd have converted all the existing files to this new format, you get more real-estate on the screen. But doing so might generate tons of conflicts since we have multiple people contributing at this time.
So my compromise was to do it only for new files, so they are future proof, and maybe later we can have a PR that updates the existing files once we're done with our batch of modifications.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs
baywet: each segment can have deprecation information, which means a path can have multiple deprecations. Then the path will take the uppermost (shortest path) deprecation.

baywet: a good example of that is `me/mailFolders/{id}/messages/{id}/attachments` where in theory:

- the user type (or a parent type) could be deprecated
- the me singleton could be depracated
- the mailFolder (or a parent type) could be deprecated
- the mailFolders navigation property could be deprecated
- ...

Assuming we have deprecation for all of those, the deprecation for the path that will be used, is the one from user, if user is not deprecated, the one from me, if not me, mailfolder, etc....

src/Microsoft.OpenApi.OData.Reader/Edm/ODataDollarCountSegment.cs
baywet: The reason why I didn't make it virtual is because I want the compiler to error out in case we forgot a segment type instead of silently pass. This is going to come handy as we coordinate with PRs that add other segment types. But it could mean a breaking change for people whom inherited from the OData segment class, so I'm willing to compromise on this one. What do you think?

xuzhg:
:shipit:  Look good to me.

PR #156, opened 2021-12-15
adds descriptions for path items

### LINKED ISSUE #154: OpenAPI document has no endpoint descriptions
The OpenAPI document generated from the microsoft graph API has descriptions for HTTP actions (e.g. GET, POST) but no descriptions for the actual endpoints. The specific use case where these descriptions would be useful is when generating a CLI from the OpenAPI file using kiota. The help docs for CLI commands would come from the descriptions.

Example:
Given an endpoint /users/messages with an action GET, the CLI would have the command structure `[cli] users messages list`. Users could then get help descriptions when they call `[cli] users messages list -h` but not `[cli] users -h` or `[cli] users messages -h`.

baywet:
fixes #154 a bug where path items would be missing descriptions

> Note: #149 will need to be updated to add a path item description as well once merged

### DIFF
--- a/test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Basic.OpenApi.V2.json
+++ b/test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Basic.OpenApi.V2.json
@@ -130,7 +130,8 @@
           }
         },
         "x-ms-docs-operation-type": "operation"
-      }
+      },
+      "x-description": "Provides operations to manage the collection of City entities."
     },
     "/City({Name})": {
       "get": {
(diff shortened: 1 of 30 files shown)


test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Basic.OpenApi.V2.json
```diff
@@ -130,7 +130,8 @@
           }
         },
         "x-ms-docs-operation-type": "operation"
-      }
+      },
+      "x-description": "Provides operations to manage the collection of City entities."
```
xuzhg, line 134: path item description is an extension?

xuzhg:
It looks good to me; :shipit:

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Basic.OpenApi.V2.json
baywet: In v2 it is, v3 it's native

PR #157, opened 2021-12-16
makes collection responses reusable

### LINKED ISSUE #125: optimization - use components for collection responses
Today each OData collection response (Collection(X)), the conversion process is inlining the response schema for each operation (object with value + next link...)
However through Microsoft Graph, we do have a lot of endpoints returning a collection of the same thing:
Collection(Message) is returned by `/me/messages`,  `/me/messageFolers/{id}/messages`, the same endpoints under `/users` and `/groups`....
Inlining those schemas results in a lot of duplication, longer descriptions and duplicated classes for code generators.
The conversion process should be declaring components instead and re-using them through multiple endpoints.

baywet:
fixes #125

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
@@ -77,6 +77,9 @@ namespace Microsoft.OpenApi.OData.Generator
                                         .Where(x => !responses.ContainsKey(x.Key)))
                             .ToDictionary(x => x.Key, x => x.Value);
 
+            if(context.HasAnyNonContainedCollections())                                        
+                responses[$"String{Constants.CollectionSchemaSuffix}"] = CreateCollectionResponse("String");
+
             return responses;
         }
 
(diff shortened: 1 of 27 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
```diff
@@ -62,10 +62,25 @@ public static OpenApiResponse GetResponse(this string statusCode)
...
             Utils.CheckArgumentNull(context, nameof(context));
 
-            return new Dictionary<string, OpenApiResponse>
+            var responses =  new Dictionary<string, OpenApiResponse>
             {
                 { "error", CreateErrorResponse() }
             };
+
+            if(context.Settings.EnableDollarCountPath)
+                responses[Constants.DollarCountSchemaName] = CreateCountResponse();
```
xuzhg, line 71: add { } to make the codes readable.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
```diff
@@ -165,6 +180,61 @@ public static OpenApiResponses CreateResponses(this ODataContext context, IEdmOp
             return responses;
         }
 
+        private static OpenApiResponse CreateCollectionResponse(IEdmStructuredType structuredType)
+        {
+            var entityType = structuredType as IEdmEntityType;
+            return CreateCollectionResponse(entityType?.FullName() ?? structuredType.FullTypeName());
+        }
+        private static OpenApiResponse CreateCollectionResponse(string typeName)
```
xuzhg, line 188: It could be remove this private static method and put the codes directly into the above?

xuzhg:
Others look good to me.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
baywet: No because this method is called twice

xuzhg:
:shipit:  Looks good to me.

PR #158, opened 2022-01-06
adds support for properties of complex type and navigation properties on complex types

### LINKED ISSUE #15: Support complex type with navigation property
Do we support the navigation property in complex type?

For example:

```
Entity Type (AddressInfo)
  |- ID (Edm.Int32)

Complex Type (Address)
    |- City (Edm.String)
    |- Street (Edm.String)
    |- AddressInfo (AddressInfo)

Entity Type (Customer)
   -> ID (Edm.Int32)
   -> Location (Address)
   -> Locations (Collection(Address))

EntitySet (Customers -> Customer)
```
So, the path:

`~/Customers/{ID}/Location/AddressInfo ` should be valid.
`~/Customers/{ID}/Locations/AddressInfo` should be invalid.

baywet:
fixes #15

This pull request adds path items for complex properties on entities as well as path items for navigation properties in complex types. Subsequently, it adds raw count, type cast and $ref path segments under those when required.
For a detail of which paths we're adding, see #15

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
@@ -3,6 +3,7 @@
 //  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // ------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OData.Edm;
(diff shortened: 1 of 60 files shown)


src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -224,6 +228,57 @@ private void RetrieveNavigationSourcePaths(IEdmNavigationSource navigationSource
...
+                    foreach (IEdmNavigationProperty np in complexTypeReference
+                                                        .DeclaredNavigationProperties()
+                                                        .Union(definition
+                                                                        .FindAllBaseTypes()
+                                                                        .SelectMany(x => x.DeclaredNavigationProperties()))
+                                                        .Distinct()
+                                                        .Where(CanFilter))
+                    {
+                        var count = _model.GetRecord<CountRestrictionsType>(np, CapabilitiesConstants.CountRestrictions);
+                        RetrieveNavigationPropertyPaths(np, count, currentPath, convertSettings);
```
peombwa, line 275: `RetrieveNavigationPropertyPaths` will throw an exception when calling [`ShouldExpandNavigationProperty`](https://github.com/microsoft/OpenAPI.NET.OData/blob/e647a1be01598b0ea9acb41325e8e47a8bf8fcaa/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L339) since[ `OdataSegment.EntityType`](https://github.com/microsoft/OpenAPI.NET.OData/blob/e647a1be01598b0ea9acb41325e8e47a8bf8fcaa/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L443) is not yet implemented:
https://github.com/microsoft/OpenAPI.NET.OData/blob/e647a1be01598b0ea9acb41325e8e47a8bf8fcaa/src/Microsoft.OpenApi.OData.Reader/Edm/ODataSegment.cs#L90

irvinesunday: Not sure if it's related but, a null exception is thrown when attempting to retrieve the "Entity type of a Complex type property segment" here:
https://github.com/microsoft/OpenAPI.NET.OData/blob/9255901b9d43b08e4a68f330bb8ace9c5bac5bad/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L443

Example:
The complex property segment in question:
image

The complex type which has containment navigation properties
image

baywet: I've just pushed another commit that should solve your issue, please pull and give it another try.

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -16,37 +17,38 @@ internal static class EdmModelHelper
...
 
             if (!derivedTypes.Any())
             {
                 return null;
             }
 
-            OpenApiSchema schema = new OpenApiSchema
-            {
+            OpenApiSchema schema = new()
+			{
```
xuzhg, line 34: revert

test/Microsoft.OpenAPI.OData.Reader.Tests/PathItem/ComplexPropertyPathItemHandlerTests.cs
```diff
@@ -0,0 +1,72 @@
+using System;
```
xuzhg, line 1: don't forget to add copyright head

xuzhg:
Basically, it looks good to me.

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
baywet: are you referring to the `new()` syntax?

test/Microsoft.OpenAPI.OData.Reader.Tests/PathItem/ComplexPropertyPathItemHandlerTests.cs
baywet: added

darrelmiller:
@baywet We cannot generate paths for properties that are complex types unless there is a specific annotation to indicate that the path is supported.  It causes way too many paths to be generated.

baywet:
@darrelmiller this PR resulted in a 3x path items increase.
https://github.com/microsoft/OpenAPI.NET.OData/pull/158/files#diff-fbd1ddb1a7f9b639e508912aeb20d5fa47fa88a3850b93aae4d81f82633cac39R51

(considering it might be impacted by inaccurate contains target attributes as well)

I'd be ok with having a default of "if there's no annotation saying it's supported, don't add anything" and this should be a quick change once we know which annotation to read in.

Which annotation should I read to implement that change?

Issue #123, opened 2021-11-10
Missing navigation path items for navigation properties with contains target


baywet:
The current beta description is missing a path item and corresponding operations for

`/teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/replies/{reply_id}/hostedContents`

Where:
- Hosted contents is a navigation property on chatMessage (no contains target)
- Replies is a collection of chatMessages navigation property on chatMessage (contains target true)
- Messages is a collection of chatMessages navigation property on channel (contains target true)
- Channels is a collection of channel navigation property on team (contains target true)
- And Teams is a collection of teams

irvinesunday:
From the `chatMessage` Entity type snippet below:

![image](https://user-images.githubusercontent.com/40403681/142479580-fb1dd8e2-2353-4db7-8092-5eeda69f8407.png)

`hostedContents` navigation property should be accessible using the path:

`/teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/hostedContents`

or to retrieve a single entity...

`/teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/hostedContents/{chatMessageHostedContent-id}`

The `replies` navigation property can be accessible using the path: 

`/teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/replies`

or to retrieve a single entity...

`/teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/replies/{chatMessage-id1}`

Expansion beyond the `.../replies/{chatMessage-id1}` path is not possible despite the containment, because the defining entity type (`graph.chatMessage`) references the same entity type (`chatMessage`) that the `replies` navigation property is contained in, and this will lead to circular references.

baywet:
Thanks for the additional information.
The expansion logic could detect circular references effectively

- not expanding /teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/replies/{reply_id}/replies, as it's already "been through" the replies navigation property
- expanding all the way to /teams/{team-id}/channels/{channel-id}/messages/{chatMessage-id}/replies/{reply_id}/hostedContents

Instead of relying on a blanket behavior of not expanding any navigation property beyond a certain point.

darrelmiller:
The problem is that replies should not be contained.  Replies are also messages and therefore their canonical path is via the messages collection.  I'll go confirm this with Anand.

darrelmiller:
[...]

PR #160, opened 2022-01-17
Updates navigation property expansion logic

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/123

This PR proposes:
- Updating the expansion logic for containment navigation properties to evaluate based on whether the navigation property  has been visited before and not the navigation property type.

NB: This approach expands on all containment navigation properties. This ends up exploding the number of paths significantly.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/EdmModelExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/EdmModelExtensions.cs
@@ -124,43 +124,7 @@ namespace Microsoft.OpenApi.OData.Edm
                 current = current.BaseEntityType();
             }
         }
-
-        /// <summary>
-        /// Checks if the <paramref name="baseType"/> is assignable to <paramref name="subtype"/>.
-        /// In other words, if <paramref name="subtype"/> is a subtype of <paramref name="baseType"/> or not.
-        /// </summary>
-        /// <param name="baseType">Type of the base type.</param>
-        /// <param name="subtype">Type of the sub type.</param>
-        /// <returns>true, if the <paramref name="baseType"/> is assignable to <paramref name="subtype"/>. Otherwise returns false.</returns>
-        public static bool IsAssignableFrom(this IEdmEntityType baseType, IEdmEntityType subtype)
-        {
-            Utils.CheckArgumentNull(baseType, nameof(baseType));
-            Utils.CheckArgumentNull(subtype, nameof(subtype));
-
-            if (baseType.TypeKind != subtype.TypeKind)
-            {
-                return false;
-            }
-
-            if (subtype.IsEquivalentTo(baseType))
-            {
-                return true;
-            }
-
-            IEdmStructuredType structuredSubType = subtype;
-            while (structuredSubType != null)
-            {
-                if (structuredSubType.IsEquivalentTo(baseType))
-                {
-                    return true;
-                }
-
-                structuredSubType = structuredSubType.BaseType;
-            }
-
-            return false;
-        }
-
+                
         /// <summary>
         /// Check whether the operation is overload in the model.
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.OpenApi.OData.Reader/Edm/EdmModelExtensions.cs
```diff
@@ -124,43 +124,7 @@ public static IEnumerable<IEdmEntityType> FindAllBaseTypes(this IEdmEntityType e
...
         }
-
-        /// <summary>
-        /// Checks if the <paramref name="baseType"/> is assignable to <paramref name="subtype"/>.
-        /// In other words, if <paramref name="subtype"/> is a subtype of <paramref name="baseType"/> or not.
-        /// </summary>
-        /// <param name="baseType">Type of the base type.</param>
-        /// <param name="subtype">Type of the sub type.</param>
-        /// <returns>true, if the <paramref name="baseType"/> is assignable to <paramref name="subtype"/>. Otherwise returns false.</returns>
-        public static bool IsAssignableFrom(this IEdmEntityType baseType, IEdmEntityType subtype)
```
baywet, line 135: since this method was public, this change is a breaking change.
Would you consider adding the method back and adding an obsolete annotation instead?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -273,15 +273,24 @@ private void RetrieveNavigationPropertyPaths(IEdmNavigationProperty navigationPr
             Debug.Assert(navigationProperty != null);
             Debug.Assert(currentPath != null);
 
+            // Check whether the navigation property has already been navigated in the path
+            foreach (ODataSegment segment in currentPath)
```
baywet, line 277: checking on the name only is not sufficient. We could have something completely different in the path with the same name.

baywet (changes requested):
does this also replace #152? 

You probably need to update the integration test files as well.

src/Microsoft.OpenApi.OData.Reader/Edm/EdmModelExtensions.cs
irvinesunday: Oh yeah.

irvinesunday:
> does this also replace #152?
> 
> You probably need to update the integration test files as well.

@baywet this PR does not replace that #152. This one just deals with the expansion logic of containment navigation properties, the other issue has to do with "turning off" containment of some navigation properties without setting `ContainsTarget=true` to `false`

baywet:
ok thanks for confirming. I believe then that #158 does what this does and more, there's just a last minute bug in there I need to address, but please have a look at the code and tell me what you think.

irvinesunday:
> ok thanks for confirming. I believe then that #158 does what this does and more, there's just a last minute bug in there I need to address, but please have a look at the code and tell me what you think.

I'm unable to run the branch on that PR due to the bug, but from a quick look at the expansion logic of containment navigation properties I don't see where it has modified the current experience.

baywet:
for sanity: this was discussed offline but this PR doesn't overlap with #160, both are needed. It however adds a lot of path items and is probably not yet ready until #152 gets resolved.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -324,24 +324,44 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
+            ODataPath currentPath,
+            OpenApiConvertSettings convertSettings,
+            Stack<string> visitedNavigationProperties = null)
         {
             Debug.Assert(navigationProperty != null);
             Debug.Assert(currentPath != null);
 
+            if (visitedNavigationProperties == null)
+            {
+                visitedNavigationProperties = new Stack<string>();
```
baywet, line 340:
```suggestion
                visitedNavigationProperties = new ();
```

baywet (approved):

baywet (approved):

xuzhg:
looks good to me.

PR #167, opened 2022-01-27
adds the ability to define error responses on error codes instead of default

### LINKED ISSUE #165: errors description should not be translating to default responses
currently the conversion lib describes the errors for any operation as a default response.
This behavior makes it really hard to do anything useful during code generation.
It should instead be using 4XX and 5XX response classes, for more information [see this thread](https://github.com/microsoft/kiota/issues/411).

Update the library to:
- [x] include a model for the standard [odata error and inner error](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ErrorResponseBody)
- [x] add a convert setting to switch between default and 4xx/5xx response class descriptions (default false)
- [x] add the responses whenever that switch is set to true

Additional information
https://docs.microsoft.com/en-us/graph/errors

baywet:
fixes #165

This PR adds a new setting `ErrorResponsesAsDefault` default `true` to avoid breaking existing users.

When true, the error responses will only be described as the following

```JSON
"responses": {
          "200": {
            "$ref": "#/components/responses/ODataCountResponse"
          },
          "default": {
            "$ref": "#/components/responses/error"
          }
        }
```

When false they'll be described as the following

```JSON
"responses": {
          "200": {
            "$ref": "#/components/responses/ODataCountResponse"
          },
          "4XX": {
            "$ref": "#/components/responses/error"
          },
          "5XX": {
            "$ref": "#/components/responses/error"
          }
        }
```

This change is made so client generators can "better understand" what the error payloads look like and generate more accurate SDKs.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
@@ -20,7 +20,8 @@ public static class OpenApiOperationExtensions
     /// <param name="operation">The operation.</param>
     /// <param name="settings">The settings.</param>
     /// <param name="addNoContent">Whether to add a 204 no content response.</param>
-    public static void AddErrorResponses(this OpenApiOperation operation, OpenApiConvertSettings settings, bool addNoContent = false) {
+    public static void AddErrorResponses(this OpenApiOperation operation, OpenApiConvertSettings settings, bool addNoContent = false)
+    {
         if (operation == null) {
             throw Error.ArgumentNull(nameof(operation));
         }
--- a/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
@@ -1,0 +1,51 @@
+// ------------------------------------------------------------
+//  Copyright (c) Microsoft Corporation.  All rights reserved.
+//  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// ------------------------------------------------------------
+
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.OData.Generator;
+
+namespace Microsoft.OpenApi.OData.Common;
+
+/// <summary>
+/// Extensions methods for the OpenApiOperation class.
+/// </summary>
+public static class OpenApiOperationExtensions
+{
+    /// <summary>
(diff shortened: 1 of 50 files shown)


src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -208,6 +208,11 @@ public string PathPrefix
         /// </summary>
         public bool EnableDeprecationInformation { get; set; } = true;
 
+        /// <summary>
+        /// Gets/sets a value indicating whether the error responses should be described as a default response or as 4XX and 5XX error responses.
```
xuzhg, line 212: I am ok to add this configuration.

src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+public static class OpenApiOperationExtensions
+{
+    /// <summary>
+    /// Adds a default response to the operation or 4XX/5XX responses for the errors depending on the settings.
+    /// Also adds a 204 no content response when requested.
+    /// </summary>
+    /// <param name="operation">The operation.</param>
+    /// <param name="settings">The settings.</param>
+    /// <param name="addNoContent">Whether to add a 204 no content response.</param>
+    public static void AddErrorResponses(this OpenApiOperation operation, OpenApiConvertSettings settings, bool addNoContent = false) {
```
xuzhg, line 23: I don't like the code style in this method.
You have the java "{}", we should keep all styles consistent.

xuzhg:
:shipit:

src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
baywet: So open curly on the next line?

baywet:
```suggestion
    public static void AddErrorResponses(this OpenApiOperation operation, OpenApiConvertSettings settings, bool addNoContent = false)
    {
```

baywet:
@xuzhg thanks for reviewing.
Holding until I get feedback on that proposal https://github.com/microsoft/OpenAPI.NET.OData/issues/165#issuecomment-1023615538

baywet:
proposal design was approved and implemented, waiting for final review before merge.

xuzhg:
Looks good to me, of course, if you can make the code alignment for better readability.
:shipit:

Issue #159, opened 2022-01-07
Path parameters should be declared on the Path Item object and not on the Operation object


baywet:
Since the path item object [supports declaring parameters](https://spec.openapis.org/oas/latest.html#path-item-object), the parameters that are in path should be declared on this object, not the operation object. This way if we have multiple operations on the same path items, the description would not have a duplicate description of the path parameter for each operation.

```diff
  '/Airports/{IcaoCode}':
    description: Provides operations to manage the collection of Airport entities.
+   parameters:
+   - name: IcaoCode
+      in: path
+      description: 'key: IcaoCode of Airport'
+      required: true
+      schema:
+        type: string
+      x-ms-docs-key-type: Airport
    get:
      tags:
        - Airports.Airport
      summary: Get entity from Airports by key
      operationId: Airports.Airport.GetAirport
      parameters:
-       - name: IcaoCode
-         in: path
-         description: 'key: IcaoCode of Airport'
-         required: true
-         schema:
-           type: string
-         x-ms-docs-key-type: Airport
        - name: $select
          in: query
          description: Select properties to be returned
          style: form
          explode: false
          schema:
            uniqueItems: true
            type: array
            items:
              enum:
                - Name
                - IcaoCode
                - IataCode
                - Location
              type: string
        - name: $expand
          in: query
          description: Expand related entities
          style: form
          explode: false
          schema:
            uniqueItems: true
            type: array
            items:
              enum:
                - '*'
              type: string
      responses:
        '200':
          description: Retrieved entity
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Microsoft.OData.Service.Sample.TrippinInMemory.Models.Airport'
        default:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
    patch:
      tags:
        - Airports.Airport
      summary: Update entity in Airports
      operationId: Airports.Airport.UpdateAirport
      parameters:
-       - name: IcaoCode
-         in: path
-         description: 'key: IcaoCode of Airport'
-         required: true
-         schema:
-           type: string
-         x-ms-docs-key-type: Airport
      requestBody:
        description: New property values
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Microsoft.OData.Service.Sample.TrippinInMemory.Models.Airport'
        required: true
      responses:
        '204':
          description: Success
        default:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
    delete:
[...]

PR #173, opened 2022-02-07
Add support for declaring path parameters on path item object instead of operation object

danielmbaluka:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/159

This PR add support for declaring path parameters on the [path item](https://spec.openapis.org/oas/latest.html#path-item-object) object instead of the [operation ](https://spec.openapis.org/oas/latest.html#operation-object) object. This feature is controlled by a new flag (`DeclarePathParametersOnPathItem`) on `OpenApiConvertSettings`

Example usage:
---
```cs
    OpenApiConvertSettings settings = new()
    {
        DeclarePathParametersOnPathItem = true,
    };
    ODataContext context = new ODataContext(model, settings);
```

Example output
---
```diff
  '/Airports/{IcaoCode}':
    description: Provides operations to manage the collection of Airport entities.
    get:
      tags:
        - Airports.Airport
      summary: Get entity from Airports by key
      operationId: Airports.Airport.GetAirport
      parameters:
-       - name: IcaoCode
-         in: path
-         description: 'key: IcaoCode of Airport'
-         required: true
-         schema:
-           type: string
-         x-ms-docs-key-type: Airport
        - name: $select
          in: query
          description: Select properties to be returned
          style: form
          explode: false
          schema:
            uniqueItems: true
            type: array
            items:
              enum:
                - Name
                - IcaoCode
                - IataCode
                - Location
              type: string
        - name: $expand
          in: query
          description: Expand related entities
          style: form
          explode: false
          schema:
            uniqueItems: true
            type: array
            items:
              enum:
                - '*'
              type: string
      responses:
        '200':
          description: Retrieved entity
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Microsoft.OData.Service.Sample.TrippinInMemory.Models.Airport'
        default:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
    patch:
      tags:
        - Airports.Airport
      summary: Update entity in Airports
      operationId: Airports.Airport.UpdateAirport
      parameters:
-       - name: IcaoCode
-         in: path
-         description: 'key: IcaoCode of Airport'
-         required: true
-         schema:
-           type: string
-         x-ms-docs-key-type: Airport
      requestBody:
        description: New property values
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Microsoft.OData.Service.Sample.TrippinInMemory.Models.Airport'
        required: true
      responses:
        '204':
          description: Success
        default:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
    delete:
      tags:
        - Airports.Airport
      summary: Delete entity from Airports
      operationId: Airports.Airport.DeleteAirport
      parameters:
-       - name: IcaoCode
-         in: path
-         description: 'key: IcaoCode of Airport'
-         required: true
-         schema:
-           type: string
-         x-ms-docs-key-type: Airport
        - name: If-Match
          in: header
          description: ETag
          schema:
            type: string
      responses:
        '204':
          description: Success
        default:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
+   parameters:
+   - name: IcaoCode
+      in: path
+      description: 'key: IcaoCode of Airport'
+      required: true
+      schema:
+        type: string
+      x-ms-docs-key-type: Airport
```

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiParameterGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiParameterGenerator.cs
@@ -217,6 +217,56 @@ namespace Microsoft.OpenApi.OData.Generator
         }
 
         /// <summary>
+        /// Creates the path parameters for the <see cref="ODataPath"/>
+        /// </summary>
+        /// <param name="path">The ODataPath</param>
+        /// <param name="context">The OData context.</param>
+        /// <returns>The created list of <see cref="OpenApiParameter"/></returns>
+        public static List<OpenApiParameter> CreatePathParameters(this ODataPath path, ODataContext context)
+        {
+            List<OpenApiParameter> pathParameters = new();
+            var parameterMappings = path.CalculateParameterMapping(context.Settings);
+
+            foreach (ODataKeySegment keySegment in path.OfType<ODataKeySegment>())
+            {
+                IDictionary<string, string> mapping = parameterMappings[keySegment];
+                pathParameters.AddRange(context.CreateKeyParameters(keySegment, mapping));
+            }
+
+            // Add the route prefix parameter v1{data}
+            if (context.Settings.RoutePathPrefixProvider?.Parameters != null)
+            {
+                pathParameters.AddRange(context.Settings.RoutePathPrefixProvider.Parameters);
+            }
+
+            return pathParameters;
+        }
+
+        /// <summary>
+        /// Adds an OpenApiParameter to an existing list of OpenApiParameters.
+        /// If a parameter with the same name already exists in the list, the new parameter name
+        /// if suffixed with an incrementing number
+        /// </summary>
+        /// <param name="parameters">The list of OpenApiParameters to be appended to</param>
+        /// <param name="parameter">The new OpenApiParameter to be appended</param>
+        public static void AppendParameter(this IList<OpenApiParameter> parameters, OpenApiParameter parameter)
+        {
+            HashSet<string> parametersSet = new(parameters.Select(p => p.Name));
+
+            string parameterName = parameter.Name;
+            int index = 1;
+            while (parametersSet.Contains(parameterName))
+            {
+                parameterName += index.ToString();
+                index++;
+            }
+
+            parameter.Name = parameterName;
+            parametersSet.Add(parameterName);
+            parameters.Add(parameter);
+        }
+
+        /// <summary>
         /// Create the $top parameter.
         /// </summary>
         /// <param name="context">The OData context.</param>
--- a/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
@@ -226,6 +226,13 @@ namespace Microsoft.OpenApi.OData
         public string InnerErrorComplexTypeName { get; set; } = "InnerError";
 
         /// <summary>
+        /// Gets/Sets a value indicating whether path parameters should be declared on path item object.
+        /// If true, path parameters will be declared on the path item object, otherwise they 
+        /// will be declared on the operation object.
+        /// </summary>
+        public bool DeclarePathParametersOnPathItem { get; set; } = false;
+
+        /// <summary>
         /// Gets/Sets a value indicating whether or not to use restrictions annotations to generate paths for complex properties.
         /// </summary>
         public bool UseRestrictionAnnotationsToGeneratePathsForComplexProperties { get; set; } = true;
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
@@ -49,7 +49,10 @@ namespace Microsoft.OpenApi.OData.Operation
                These need to be set before Responses, as the Parameters
                will be used in the Responses when creating Links.
             */
-            SetParameters(operation);
+            if (!Context.Settings.DeclarePathParametersOnPathItem)
+            {
+                SetParameters(operation);
+            }
 
             // Responses
             SetResponses(operation);
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
@@ -25,6 +25,11 @@ namespace Microsoft.OpenApi.OData.Operation
 
         protected IDictionary<ODataSegment, IDictionary<string, string>> ParameterMappings;
 
+        /// <summary>
+        /// The path parameters in the path
+        /// </summary>
+        protected IList<OpenApiParameter> PathParameters;
+
         /// <inheritdoc/>
         public virtual OpenApiOperation CreateOperation(ODataContext context, ODataPath path)
         {
@@ -139,9 +144,10 @@ namespace Microsoft.OpenApi.OData.Operation
         /// <param name="operation">The <see cref="OpenApiOperation"/>.</param>
         protected virtual void SetParameters(OpenApiOperation operation)
         {
+            PathParameters = Path.CreatePathParameters(Context);
             if (!Context.Settings.DeclarePathParametersOnPathItem)
             {
-                foreach(var parameter in Path.CreatePathParameters(Context))
+                foreach (var parameter in PathParameters)
                 {
                     operation.Parameters.AppendParameter(parameter);
                 }
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
@@ -49,10 +49,7 @@ namespace Microsoft.OpenApi.OData.Operation
                These need to be set before Responses, as the Parameters
                will be used in the Responses when creating Links.
             */
-            if (!Context.Settings.DeclarePathParametersOnPathItem)
-            {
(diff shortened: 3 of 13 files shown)


src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
```diff
@@ -101,5 +113,54 @@ protected virtual void AddOperation(OpenApiPathItem item, OperationType operatio
...
+            {
+                IDictionary<string, string> mapping = ParameterMappings[keySegment];
+                foreach (var parameter in Context.CreateKeyParameters(keySegment, mapping))
+                {
+                    AppendParameter(item, parameter);
+                }
+            }
+
+            // Add the route prefix parameter v1{data}
+            if (Context.Settings.RoutePathPrefixProvider != null && Context.Settings.RoutePathPrefixProvider.Parameters != null)
```
baywet, line 133:
```suggestion
            if (Context.Settings.RoutePathPrefixProvider?.Parameters != null)
```

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
```diff
@@ -101,5 +113,54 @@ protected virtual void AddOperation(OpenApiPathItem item, OperationType operatio
...
+            {
+                foreach (var parameter in Context.Settings.RoutePathPrefixProvider.Parameters)
+                {
+                    item.Parameters.Add(parameter);
```
baywet, line 137: don't you have an add range method that'd prevent manually looping here?

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -225,6 +225,13 @@ public string PathPrefix
         /// </summary>
         public string InnerErrorComplexTypeName { get; set; } = "InnerError";
 
+        /// <summary>
+        /// Gets/Sets a value indicating whether path parameters should be declared on path item object.
+        /// If true, path parameters will be declared on the path item object, otherwise they 
+        /// will be declared on the operation object.
+        /// </summary>
+        public bool DeclarePathParametersOnPathItem { get; set; } = false;
```
baywet, line 233: an entry needs to be added in this file for this new property https://github.com/microsoft/OpenAPI.NET.OData/blob/master/src/Microsoft.OpenApi.OData.Reader/PublicAPI.Unshipped.txt

baywet (changes requested):
Thanks for taking that, a few comments.

src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
```diff
@@ -139,12 +139,15 @@ protected virtual void SetRequestBody(OpenApiOperation operation)
         /// <param name="operation">The <see cref="OpenApiOperation"/>.</param>
         protected virtual void SetParameters(OpenApiOperation operation)
         {
-            foreach (ODataKeySegment keySegment in Path.OfType<ODataKeySegment>())
+            if (!Context.Settings.DeclarePathParametersOnPathItem)
```
irvinesunday, line 142: Why not check for this flag in line 52? `SetParameters(operation);`

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
```diff
@@ -101,5 +113,54 @@ protected virtual void AddOperation(OpenApiPathItem item, OperationType operatio
...
+                }
+            }
+        }
+
+        protected static void AppendParameter(OpenApiPathItem item, OpenApiParameter parameter)
```
irvinesunday, line 142: I believe we can refactor out this method (and the `SetParameters(OpenApiPathItem item)` method) into a common module that can be re-used by the `OperationHandler.cs` class as well.

irvinesunday:
Let's keep in mind that [OpenApiLinks](https://github.com/microsoft/OpenAPI.NET.OData/blob/7b0f3cde539e8d3ff95fce3b31276424bc519a88/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L21-L33) are currently generated based on the operation parameters:

image

We still need to be able to generate these links even when the parameters are not available at the operation level.

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
danielmbaluka: Good suggestion. I have adopted it. Thanks.

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
danielmbaluka: @baywet `Parameters` are declared with a type of `IList<T>` and not `List<T>`. `AddRange` is defined in `List<T>`, so for us to use it we need to convert the IList to a List e.g. `item.Parameters.ToList().AddRange(...)`. The conversion uses the same `foreach` loop internally, so we don't get any performance gain. The resulting code is more concise though.

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
danielmbaluka: Yes. I am working on it.

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
baywet: ToList() is noop if the backing type is in effect a List if I remember things correctly, you'd have to reassign the properly after doing this though, your call.

src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
danielmbaluka: The reason for doing it here is because `SetParameters(operation)` in `OperationHandler` has additional functionality to create more query and header parameters depending on the operation e.g. $top, $filter etc https://github.com/microsoft/OpenAPI.NET.OData/blob/master/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs#L151-L160. This is unlike in `PathItemHandler` where we are only interested in path parameters. So the flag should only be applied to control whether we add path parameters in OperationHandler or not, and not affect the query and header parameters.

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
danielmbaluka: This has been done, thanks.

src/Microsoft.OpenApi.OData.Reader/PathItem/PathItemHandler.cs
danielmbaluka: @irvinesunday the shared code has been refactored into a common module and re-used for both `OperationHandler.cs` and `PathItemHandler.cs`

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiParameterGenerator.cs
```diff
@@ -216,6 +216,56 @@ public static IList<OpenApiParameter> CreateParameters(this ODataContext context
...
+                pathParameters.AddRange(context.CreateKeyParameters(keySegment, mapping));
+            }
+
+            // Add the route prefix parameter v1{data}
+            if (context.Settings.RoutePathPrefixProvider?.Parameters != null)
+            {
+                pathParameters.AddRange(context.Settings.RoutePathPrefixProvider.Parameters);
+            }
+
+            return pathParameters;
```
baywet, line 242: shouldn't this also add the odata functions parameters? e.g. `/GetSalesTaxRate(PostalCode=10)`

danielmbaluka:
> Let's keep in mind that [OpenApiLinks](https://github.com/microsoft/OpenAPI.NET.OData/blob/7b0f3cde539e8d3ff95fce3b31276424bc519a88/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiLinkGenerator.cs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L21-L33) are currently generated based on the operation parameters:
> 
> image
> 
> We still need to be able to generate these links even when the parameters are not available at the operation level.

Thanks @irvinesunday for catching this (and the suggestion on how to handle it). I have pushed a fix for to resolve it https://github.com/microsoft/OpenAPI.NET.OData/pull/173/commits/c1365164496a64476245bc3882052ae080fc96a7

danielmbaluka:
@baywet @irvinesunday  @xuzhg @darrelmiller kindly have a look at the PR.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiParameterGenerator.cs
darrelmiller: There is no benefit of using Path Parameters for OData Actions and Functions.  Even composable functions.

baywet (approved):

irvinesunday (approved):

PR #174, opened 2022-02-07
errors naming conventions improvements

### LINKED ISSUE #172: Error conversion adjustments
1. The main error types live under a different hardcoded namespace instead of the main one we're generating to. This makes for ugly clients and we should change that
1. The main error type is named error, which conflicts with the native error type in typescript and Go. And because there's also a property named error on that type, it also conflicts in C#. We should change that as well.

baywet:
- fixes #172 
- moved error definitions under main namespace
- renames types to avoid conflicts with native types or naming conventions

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiErrorSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiErrorSchemaGenerator.cs
@@ -17,6 +17,11 @@ namespace Microsoft.OpenApi.OData.Generator
     /// </summary>
     internal static class OpenApiErrorSchemaGenerator
     {
+        internal const string ODataErrorClassName = "ODataError";
+        internal const string MainErrorClassName = "MainError";
+        internal const string ErrorDetailsClassName = "ErrorDetails";
+        internal const string InnerErrorClassName = "InnerError";
+
         /// <summary>
         /// Create the dictionary of <see cref="OpenApiSchema"/> object.
         /// The name of each pair is the namespace-qualified name of the type. It uses the namespace instead of the alias.
@@ -31,16 +36,12 @@ namespace Microsoft.OpenApi.OData.Generator
 
             return new Dictionary<string, OpenApiSchema>()
             {
-                { $"{rootNamespaceName}{OdataErrorClassName}", CreateErrorSchema(rootNamespaceName) },
+                { $"{rootNamespaceName}{ODataErrorClassName}", CreateErrorSchema(rootNamespaceName) },
                 { $"{rootNamespaceName}{MainErrorClassName}", CreateErrorMainSchema(rootNamespaceName) },
                 { $"{rootNamespaceName}{ErrorDetailsClassName}", CreateErrorDetailSchema() },
                 { $"{rootNamespaceName}{InnerErrorClassName}", CreateInnerErrorSchema(context) }
             };
         }
-        internal const string OdataErrorClassName = "ODataError";
-        internal const string MainErrorClassName = "MainError";
-        internal const string ErrorDetailsClassName = "ErrorDetails";
-        internal const string InnerErrorClassName = "InnerError";
 
         /// <summary>
         /// Gets the error namespace name based on the root namespace of the model.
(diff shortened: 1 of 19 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiErrorSchemaGenerator.cs
```diff
@@ -27,28 +27,36 @@ internal static class OpenApiErrorSchemaGenerator
...
-
-                // odata.error.main
-                { "odata.error.main", CreateErrorMainSchema() },
-
-                // odata.error.detail
-                { "odata.error.detail", CreateErrorDetailSchema() },
-
-                // odata.error.innererror
-                { "odata.error.innererror", CreateInnerErrorSchema(context) }
+                { $"{rootNamespaceName}odataerror", CreateErrorSchema(rootNamespaceName) },
```
irvinesunday, line 34: Can we define as `const` these string values that have been used in more than one place so that they are easily re-usable within the doc.?

baywet: do you mean the odataerror part?

baywet: we should probably upper camel case these so we get proper class names in the code-gen too.

irvinesunday: > do you mean the odataerror part?

Yes, and the `mainerror`, `errordetails`, etc...

baywet: thanks, solved

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiErrorSchemaGenerator.cs
```diff
@@ -27,28 +27,40 @@ internal static class OpenApiErrorSchemaGenerator
...
-
-                // odata.error.innererror
-                { "odata.error.innererror", CreateInnerErrorSchema(context) }
+                { $"{rootNamespaceName}{OdataErrorClassName}", CreateErrorSchema(rootNamespaceName) },
+                { $"{rootNamespaceName}{MainErrorClassName}", CreateErrorMainSchema(rootNamespaceName) },
+                { $"{rootNamespaceName}{ErrorDetailsClassName}", CreateErrorDetailSchema() },
+                { $"{rootNamespaceName}{InnerErrorClassName}", CreateInnerErrorSchema(context) }
             };
         }
+        internal const string OdataErrorClassName = "ODataError";
```
irvinesunday, line 40: Out of curiosity: Why did you choose to define the `const` values below where they are first used?

baywet: because it doesn't matter to the compiler, or to me. I use symbol navigation rather than scrolling when I browse code (either f12 or ctrl + P and then type @ in vs code)

irvinesunday: I think for uniformity purposes (see [example](https://github.com/microsoft/OpenAPI.NET.OData/blob/f72545004b24a650dfc1d885d0e85bd4656d22ff/src/Microsoft.OpenApi.OData.Reader/Operation/ODataTypeCastGetOperationHandler.cs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L43-L51)) with other sections of the code base and for easier reading (by other maintainers) you could stick to defining these before they are first used?

baywet: changed. Give a try to symbol navigation instead of scrolling, you'll see it's much faster!

irvinesunday: Thank you for taking in this suggestion.

irvinesunday (approved):

Issue #177, opened 2022-02-14
Use descriptions from restrictions annotations for operation descriptions


irvinesunday:
We should use the restrictions capability annotations in the model for the operations descriptions. For example, given the below restrictions annotations for the `User` entity type:

![MicrosoftTeams-image (30)](https://user-images.githubusercontent.com/40403681/153857983-b4221e9b-95df-4297-ab2d-418ead119fa7.png)

In the converted OpenAPI document, the operation description would be emitted as below:
![MicrosoftTeams-image (31)](https://user-images.githubusercontent.com/40403681/153858007-31c5fd7e-ffe4-4d5f-9ee0-89df10612b1f.png)

_The green highlights indicate descriptions retrieved from the restrictions annotations ._

maisarissi:
This feature is needed as part of to the https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/876

Issue #176, opened 2022-02-10
only add complextype property path items when CRUD restrictions are present


baywet:
#158 added additional paths for the structured properties (complex types) and their navigation properties. This resulted in an important growth of the number of path items being generated because this is enabled by default, and it doesn't read any annotation.

We should change the behavior to add the path items for complex properties only if readrestrictions/updaterestrictions (single properties or collections), create restrictions (collections) are present. The subsequent navigation properties paths for the complex type are only added when the parent structured/complex property is inserted.

We're not considering delete restrictions -> One can achieve the same thing by updating to null, and can't address single items for collections, guidance is to use entities at this point so it shouldn't be present.

We should also add a new convert setting "require CRUD restrictions to add structured properties path items", default true, so people can have the path items injected without having to add the annotations should they want to (when they set the setting to false).

code pointers
https://github.com/microsoft/OpenAPI.NET.OData/blob/42943815b117c3420795f32904953dca98df5541/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L246

TODO:
- [x] remove delete operations code for complex properties.
- [x] new convert setting "require CRUD restrictions to add structured properties path items", default true.
- [x] Only insert GET operation for complex properties if the new setting is false or if a read restriction is present
- [x] Only insert PATCH operation for complex properties if the new setting is false or if a update restriction is present
- [x] Only insert POST operation for complex properties when the property is a collection if the new setting is false or if a create restriction is present
- [x] update trip service (CSDL and output) to contain test data
- [x] update unit tests

PR #178, opened 2022-02-18
Uses descriptions from CRUD restrictions annotations for OpenAPI operation descriptions

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/177

This PR proposes:
- Reading in descriptions for operations from the below CRUD restrictions annotations:
  - `Org.OData.Capabilities.V1.ReadRestrictions`
  - `Org.OData.Capabilities.V1.UpdateRestrictions`
  - `Org.OData.Capabilities.V1.InsertRestrictions`
  - `Org.OData.Capabilities.V1.DeleteRestrictions`
  
  For `Get` operations, if the above are unavailable, we fallback to the `Org.OData.Core.V1.Description` annotation for that particular EDM element. This is the standard annotation that we've been using for our descriptions, which we've used for all operations. We will use this default description annotation only for `Get` operations as they don't really make sense for other operations since they are mostly just descriptive of a given EDM element and don't really provide any meaningful info. about a particular operation. 
 Since `Org.OData.Capabilities.V1.NavigationRestrictions` can be defined at the navigation source (entity set/singleton) level as well as in-lined at the navigation property level, we are checking for descriptions at the aforementioned annotatable targets when populating the operation descriptions for navigation properties and `$ref` paths.

- Updating tests and integration files to validate the above.
- Updating the checks for navigation property `Navigability` to check for `Org.OData.Capabilities.V1.NavigationRestrictions` at both the navigation source level and navigation property level.
- Refactoring out some common code into a common helper or utility class for reuse.

_Open question_: Is there a way to annotate structural/complex properties with CRUD restrictions annotations? In the absence of this, we aren't able to populate meaningful descriptions for complex properties.


src/Microsoft.OpenApi.OData.Reader/Operation/ComplexPropertyDeleteOperationHandler.cs
```diff
@@ -24,9 +24,6 @@ protected override void SetBasicInfo(OpenApiOperation operation)
         // Summary
         operation.Summary = $"Delete {ComplexPropertySegment.Property.Name} property value";
 
-        // Description
-        operation.Description = Context.Model.GetDescriptionAnnotation(ComplexPropertySegment.Property);
```
baywet, line 28: I'm confused to why we're removing the description for those operations?

src/Microsoft.OpenApi.OData.Reader/Operation/EntitySetGetOperationHandler.cs
```diff
@@ -153,31 +165,29 @@ protected override void SetResponses(OpenApiOperation operation)
 
         protected override void SetSecurity(OpenApiOperation operation)
         {
-            ReadRestrictionsType read = Context.Model.GetRecord<ReadRestrictionsType>(EntitySet, CapabilitiesConstants.ReadRestrictions);
-            if (read == null || read.Permissions == null)
+            if (ReadRestrictions == null || ReadRestrictions.Permissions == null)
```
baywet, line 168:
```suggestion
            if (ReadRestrictions?.Permissions == null)
```

src/Microsoft.OpenApi.OData.Reader/Operation/EntitySetPostOperationHandler.cs
```diff
@@ -77,31 +90,29 @@ protected override void SetResponses(OpenApiOperation operation)
 
         protected override void SetSecurity(OpenApiOperation operation)
         {
-            InsertRestrictionsType insert = Context.Model.GetRecord<InsertRestrictionsType>(EntitySet, CapabilitiesConstants.InsertRestrictions);
-            if (insert == null || insert.Permissions == null)
+            if (InsertRestrictions == null || InsertRestrictions.Permissions == null)
```
baywet, line 93:
```suggestion
            if (InsertRestrictions?.Permissions == null)
```

src/Microsoft.OpenApi.OData.Reader/Operation/EntityUpdateOperationHandler.cs
```diff
@@ -88,31 +100,29 @@ protected override void SetResponses(OpenApiOperation operation)
 
         protected override void SetSecurity(OpenApiOperation operation)
         {
-            UpdateRestrictionsType update = Context.Model.GetRecord<UpdateRestrictionsType>(EntitySet, CapabilitiesConstants.UpdateRestrictions);
-            if (update == null || update.Permissions == null)
+            if (UpdateRestrictions == null || UpdateRestrictions.Permissions == null)
```
baywet, line 103:
```suggestion
            if (UpdateRestrictions?.Permissions == null)
```

src/Microsoft.OpenApi.OData.Reader/Operation/RefGetOperationHandler.cs
```diff
@@ -40,7 +40,11 @@ protected override void SetBasicInfo(OpenApiOperation operation)
                 operation.OperationId = GetOperationId(prefix);
             }
 
-            base.SetBasicInfo(operation);
+            // Description
+            ReadRestrictionsType readRestriction = Restriction?.ReadRestrictions;
+            operation.Description = readRestriction != null
+                ? LastSegmentIsKeySegment ? readRestriction.ReadByKeyRestrictions?.Description : readRestriction.Description
+                : Context.Model.GetDescriptionAnnotation(NavigationProperty);
```
baywet, line 47:
```suggestion
            operation.Description = (LastSegmentIsKeySegment ? readRestriction?.ReadByKeyRestrictions?.Description : readRestriction?.Description) ?? Context.Model.GetDescriptionAnnotation(NavigationProperty);
```

src/Microsoft.OpenApi.OData.Reader/Operation/SingletonGetOperationHandler.cs
```diff
@@ -120,32 +133,30 @@ protected override void SetResponses(OpenApiOperation operation)
         /// <inheritdoc/>
         protected override void SetSecurity(OpenApiOperation operation)
         {
-            ReadRestrictionsType read = Context.Model.GetRecord<ReadRestrictionsType>(Singleton, CapabilitiesConstants.ReadRestrictions);
-            if (read == null || read.Permissions == null)
+            if (ReadRestrictions == null || ReadRestrictions.Permissions == null)
```
baywet, line 136:
```suggestion
            if (ReadRestrictions?.Permissions == null)
```

src/Microsoft.OpenApi.OData.Reader/Operation/SingletonPatchOperationHandler.cs
```diff
@@ -91,32 +104,30 @@ protected override void SetResponses(OpenApiOperation operation)
         /// <inheritdoc/>
         protected override void SetSecurity(OpenApiOperation operation)
         {
-            UpdateRestrictionsType update = Context.Model.GetRecord<UpdateRestrictionsType>(Singleton, CapabilitiesConstants.UpdateRestrictions);
-            if (update == null || update.Permissions == null)
+            if (UpdateRestrictions == null || UpdateRestrictions.Permissions == null)
```
baywet, line 107:
```suggestion
            if (UpdateRestrictions?.Permissions == null)
```

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -61,6 +62,36 @@ internal static OpenApiSchema GetDerivedTypesReferenceSchema(IEdmStructuredType
...
+            NavigationRestrictionsType restrictionType,
+            NavigationPropertyRestriction restrictionProperty)
+        {
+            // Verify using individual navigation restriction first
+            if (restrictionProperty != null && restrictionProperty.Navigability != null && restrictionProperty.Navigability.Value == NavigationType.None)
+            {
+                return false;
+            }
+
+            if (restrictionProperty == null || restrictionProperty.Navigability == null)
```
baywet, line 83:
```suggestion
            if (restrictionProperty?.Navigability == null)
```

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -348,25 +348,36 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
 
-            // append a navigation property.
+            // Append a navigation property.
             currentPath.Push(new ODataNavigationPropertySegment(navigationProperty));
             AppendPath(currentPath.Clone());
             visitedNavigationProperties.Push(navPropFullyQualifiedName);
 
             // Check whether a collection-valued navigation property should be indexed by key value(s).
-            NavigationPropertyRestriction restriction = navigation?.RestrictedProperties?.FirstOrDefault();
+            bool? indexableByKey = restriction?.IndexableByKey;
```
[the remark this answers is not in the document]
irvinesunday: Thanks for this improvement. 
Actually we need to set it to `true` if value evaluates to `null`. 
Also, I'm retaining the explicit typing `bool` for uniformity with other sections of the code base.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -348,25 +348,36 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
 
-            if (restriction == null || restriction.IndexableByKey == true)
+            if (indexableByKey == null || indexableByKey == true)
```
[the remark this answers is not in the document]
irvinesunday: 👍🏽 this suggestion has been applied locally.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -348,25 +348,36 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
                 return;
             }
 
+            // Get the navigation source for this navigation property
+            IEdmVocabularyAnnotatable annotatableNavigationSource = (IEdmVocabularyAnnotatable)(currentPath.FirstSegment as ODataNavigationSourceSegment).NavigationSource;
```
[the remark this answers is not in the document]
irvinesunday: Thanks for the improvement suggestion.
I'm however retaining the explicit typing `IEdmVocabularyAnnotatable` for uniformity with other sections of the code base.

src/Microsoft.OpenApi.OData.Reader/Operation/ComplexPropertyDeleteOperationHandler.cs
irvinesunday: If you check the [diff](https://github.com/microsoft/OpenAPI.NET.OData/pull/178/files#diff-dcf875e9d85634aeb21f747c8f18c1aa7d1372eb3ee45e275eedfe0a6325df53) for the _TripService.OpenApi_ file and inspect the original operations descriptions for the `Delete`, `Update` and `Post` operations, you will notice that the text is somewhat misleading with regard to the operation's intention and the operation summary text. These generic default descriptions only seem to make sense for `Get` operations.
This is my subjective viewpoint.

baywet: Understood, thanks for the additional context. If the library is not reading the right information today, shouldn't it try to read better information from somewhere else instead of not reading anything at all? What I'm worried about is that after shipping this we get an issue to have descriptions for those operations. Alternatively we could still place a generic description, but a bit more explicit.

darrelmiller:
@irvinesunday Can you confirm that this code is dependent on the ongoing work in ApiDoctor https://github.com/OneDrive/apidoctor/pull/116 that isn't yet merged?

src/Microsoft.OpenApi.OData.Reader/Operation/ComplexPropertyDeleteOperationHandler.cs
irvinesunday: For the descriptions of operations of complex properties, I propose we use the descriptions from the CRUD restriction annotations targeting complex properties. We are already going to target complex properties with CRUD restrictions annotations in the fix that addresses this issue [only add complextype property path items when CRUD restrictions are present](https://github.com/microsoft/OpenAPI.NET.OData/issues/176) anyway.

But however, for non-GET operations, I propose we still don't use the default fallback description: `Context.Model.GetDescriptionAnnotation(ComplexPropertySegment.Property)`. This way, we provide an avenue for adding CRUD-specific descriptions for complex properties and avoid outputting  default descriptions that are somewhat misleading with regard to the intent of the particular operation.

irvinesunday:
> @irvinesunday Can you confirm that this code is dependent on the ongoing work in ApiDoctor [OneDrive/apidoctor#116](https://github.com/OneDrive/apidoctor/pull/116) that isn't yet merged?

@darrelmiller they are related but independent work items. This code checks for the existence of said CRUD restriction annotations in the CSDL that the code in Api Doctor will add in that mentioned ongoing work. The structure of those annotations is already pre-determined. The absence of descriptions in the CRUD restrictions annotations, or absence of the CRUD restriction annotations themselves in the CSDL should not break the OpenAPI generation process (other than the missing operation descriptions that are being targeted). Therefore, this PR, once reviewed and validated should be able to be merged whilst the Api Doctor work is still in progress. And vice-versa I believe.

src/Microsoft.OpenApi.OData.Reader/Operation/ComplexPropertyDeleteOperationHandler.cs
baywet: couldn't we have both? get the description from the restriction annotation that matches the operation when present, and if we don't find anything, revert back to the property description with some formatting?

irvinesunday: What sort of formatting?

baywet: `Gets the <PropertyName> value for the <ParentTypeName>.` or something along those lines?

xuzhg:
Looks good to me, :shipit:

src/Microsoft.OpenApi.OData.Reader/Operation/ComplexPropertyDeleteOperationHandler.cs
irvinesunday: That feels similar to what we have in the operation `summary` texts.
@darrelmiller do you have any opinions with regard to this?

darrelmiller: The summary text is fairly clear.  I agree that putting the model property description as the operation description is confusing.  Once we have the restriction annotations on complex types, then we could pull the description from there.  Having both the summary and description mechanically generated doesn't seem to add value in my opinion.

darrelmiller (approved):

darrelmiller:
@irvinesunday Yes, the description fields are well defined. However, another element of the ApiDoctor work was to include the URL to the reference documentation.  The way we incorporate that into the CSDL has not been determined.  I guess we can add support for that in a separate set of PRs.

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -61,6 +62,36 @@ internal static OpenApiSchema GetDerivedTypesReferenceSchema(IEdmStructuredType
...
+        /// </summary>
+        /// <param name="restrictionType">The <see cref="NavigationRestrictionsType"/>.</param>
+        /// <param name="restrictionProperty">The <see cref="NavigationPropertyRestriction"/>.</param>
+        /// <returns></returns>
+        internal static bool NavigationRestrictionsAllowsNavigability(
+            NavigationRestrictionsType restrictionType,
+            NavigationPropertyRestriction restrictionProperty)
+        {
+            // Verify using individual navigation restriction first
+            if (restrictionProperty != null && restrictionProperty.Navigability != null && restrictionProperty.Navigability.Value == NavigationType.None)
```
baywet, line 78:
```suggestion
            if (restrictionProperty?.Navigability?.Value == NavigationType.None)
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -360,36 +360,47 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
+            string navPropFullyQualifiedName = $"{navigationProperty.DeclaringType.FullTypeName()}/{navigationProperty.Name}";
 
             // Check whether the navigation property has already been navigated in the path
             if (visitedNavigationProperties.Contains(navPropFullyQualifiedName))
             {
                 return;
             }
 
+            // Get the navigation source for this navigation property
+            IEdmVocabularyAnnotatable annotatableNavigationSource = (currentPath.FirstSegment as ODataNavigationSourceSegment).NavigationSource as IEdmVocabularyAnnotatable;
```
baywet, line 375:
```suggestion
            IEdmVocabularyAnnotatable annotatableNavigationSource = (currentPath.FirstSegment as ODataNavigationSourceSegment)?.NavigationSource as IEdmVocabularyAnnotatable;
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -360,36 +360,47 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
+            NavigationRestrictionsType navPropRestrictionType = _model.GetRecord<NavigationRestrictionsType>(navigationProperty, CapabilitiesConstants.NavigationRestrictions);
+            
+            NavigationPropertyRestriction restriction = navSourceRestrictionType?.RestrictedProperties?
+                .FirstOrDefault(r => r.NavigationProperty == currentPath.NavigationPropertyPath(navigationProperty.Name))
+                ?? navPropRestrictionType?.RestrictedProperties?.FirstOrDefault();
+
             // Check whether the navigation property should be part of the path
-            NavigationRestrictionsType navigation = _model.GetRecord<NavigationRestrictionsType>(navigationProperty, CapabilitiesConstants.NavigationRestrictions);
-            if (navigation != null && !navigation.IsNavigable)
+            if (EdmModelHelper.NavigationRestrictionsAllowsNavigability(navSourceRestrictionType, restriction) == false ||
```
baywet, line 386:
```suggestion
            if (!EdmModelHelper.NavigationRestrictionsAllowsNavigability(navSourceRestrictionType, restriction) ||
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -360,36 +360,47 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
+                EdmModelHelper.NavigationRestrictionsAllowsNavigability(navPropRestrictionType, restriction) == false)
```
baywet, line 387:
```suggestion
                !EdmModelHelper.NavigationRestrictionsAllowsNavigability(navPropRestrictionType, restriction))
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -409,7 +420,8 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
                 // ~/entityset/{key}/single-valued-Nav/subtype
                 CreateTypeCastPaths(currentPath, convertSettings, navigationProperty.DeclaringType, navigationProperty, targetsMany);
 
-                if (navigation?.Referenceable == true)
+                if (navSourceRestrictionType?.Referenceable == true ||
```
baywet, line 423:
```suggestion
                if (navSourceRestrictionType?.Referenceable ?? false ||
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -409,7 +420,8 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
+                    navPropRestrictionType?.Referenceable == true)
```
baywet, line 424:
```suggestion
                    navPropRestrictionType?.Referenceable ?? false)
```

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -61,6 +62,36 @@ internal static OpenApiSchema GetDerivedTypesReferenceSchema(IEdmStructuredType
...
+                // if the individual has no navigability setting, use the global navigability setting
+                if (restrictionType != null && !restrictionType.IsNavigable)
+                {
+                    // Default navigability for all navigation properties of the annotation target.
+                    // Individual navigation properties can override this value via `RestrictedProperties/Navigability`.
+                    return false;
+                }
+            }
+
+            return true;
```
baywet, line 94:
```suggestion
            return restrictionProperty?.Navigability != null || restrictionType == null || restrictionType.IsNavigable;
            // if the individual has no navigability setting, use the global navigability setting
            // Default navigability for all navigation properties of the annotation target.
            // Individual navigation properties can override this value via `RestrictedProperties/Navigability`.
```

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: The `FirstSegment` of the path will **always** be a navigation source segment, no?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: This suggestion throws the error:

image

baywet: so maybe this instead

```CSharp
if ((navSourceRestrictionType?.Referenceable ?? false) ||
                    (navPropRestrictionType?.Referenceable ?? false))
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
baywet: Functionally, maybe, but somebody would have to read all the upstream code to make sure of that, and a change in the upstream code might have impact here. Semantically if first segment is something else, or null, the app will get a null value here and run into a null ref ex, which should be guarded against.

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
irvinesunday: This suggestion throws the error:

image

Perhaps?
`if (restrictionProperty?.Navigability != null && restrictionProperty.Navigability.Value == NavigationType.None)`

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
baywet: I'm nitpicking at this point but semantically they don't mean the same thing even though they achieve the same result. ?? false goes hand in hand with the null propagation operator to say "if you don't have a value for the navSourceRestrictionType, use false as a default". Whereas == true does an implicit cast from bool? to bool

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: Okay. Got it! 
Consequently, the `NavigationRestrictionType` variable assignments will have to go into a `null` condition check then.

image

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: That works

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: Understood. Resolved.

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
baywet: yep, go with that, I thought the null propagation + implicit cast was smarter than that but apparently it's not.

baywet (approved):
Thanks for taking the time to make the changes

PR #181, opened 2022-02-23
Renames setting property

irvinesunday:
This PR:
- Renames the setting property `UseRestrictionAnnotationsToGeneratePathsForComplexProperties` to `UseRestrictionAnnotationsToGenerateComplexPropertyPaths`. The new name is more terse.


baywet (changes requested):
Instead of use may I suggest require? This maps closer to the behavior

baywet (approved):

PR #184, opened 2022-02-24
pipelines alignment

baywet:
for reviewers [here is a successful publishing run](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=68636&view=results) that relies on environments for approval (latest recommendation from ADO) instead of classic releases (getting deprecated).

the result being we have [a preview version that has been published](https://www.nuget.org/packages/Microsoft.OpenApi.OData/1.0.10-preview1)


.azure-pipelines/ci-build.yml
```diff
@@ -0,0 +1,206 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+name: $(BuildDefinitionName)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)
+
+trigger:
+  branches:
+    include:
+      - master
```
irvinesunday, line 9: OpenAPI.NET uses `vnext` branch as a default branch before `master`. We may want to have this flow in this repo as well.

baywet: I'd prefer if we aligned all of them on main like the other repos we have at a later date if you don't mind?

irvinesunday: Leaving this as is will mean all merges to `master` will do a release. And this is the main branch for this repo.

PR #194, opened 2022-03-03
fixes a bug where switch to semver versions would make comparison script fail

baywet:
once we deployed preview1 to nuget, this switched the package to semver mode, and to a different API endpoint.
This PR fixes the version comparison script.


danielmbaluka (approved):
Looks good @baywet . FYI I had to install version 1 of **SemVerPS** in order to test the script locally. See https://github.com/Sarafian/SemVerPS/#core-editions-only

baywet:
thanks for the insight, that's probably because you're running on powershell classic (as opposed to PS core). 
Made a quick commit to instruct the task to use ps core

Issue #105, opened 2021-06-10
Add support for optional parameters for operations


dagirard:
OData operations (action/functions) can have optional parameters that can be omitted from the path, and instead be specified as query parameter.

Specification: http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_InlineParameterSyntax

Example:
GET http://host/service/EmployeesByManager?ManagerID=3

ManagementID parameter has the annotation: "Core.OptionalParameter".

Today the library marks all the operations parameters as "in: path", when in fact optional parameters could be marked as "in: query", and they can be omitted.

### Assemblies affected

OpenAPI.NET.OData

### Steps to reproduce

ODataConventionModelBuilder:
ParameterConfiguration parameterConfig = operation.Parameter(int, "ManagerId");
parameterConfig.Optional().HasDefaultValue("10");

### Expected result

ManagerId parameter is optional, and identified in the OpenAPI document as a query parameter.

### Actual result

ManagerId is not optional and part of the path in the OpenAPI document.

irvinesunday:
There's a missing functionality in the Edm library. I have created an [issue](https://github.com/OData/odata.net/issues/2337) for it in their repo and asked them to action it as a P1.

PR #195, opened 2022-03-10
Adds support for optional parameters for operations

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/105

- `IEdmOperationParameter`s that have the annotation `Org.OData.Core.V1.OptionalParameter` implement `IEdmOptionalParameter` and can be marked as `in: query` and have their `required` property set to `false`.
- Adds test to validate the above.


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiParameterGenerator.cs
```diff
@@ -128,8 +127,8 @@ public static IList<OpenApiParameter> CreateParameters(this ODataContext context
                     parameter = new OpenApiParameter
                     {
                         Name = parameterNameMapping == null ? edmParameter.Name : parameterNameMapping[edmParameter.Name],
-                        In = ParameterLocation.Path,
-                        Required = true,
+                        In = edmParameter is IEdmOptionalParameter ? ParameterLocation.Query : ParameterLocation.Path,
+                        Required = edmParameter is not IEdmOptionalParameter,
```
baywet, line 131: as one is the inverse of the other, just assign the result to a boolean var above, and use that boolean instead in both places

baywet (changes requested):

baywet (approved):

Issue #202, opened 2022-03-25
Incorrect paths for functions with structured or collection-valued parameters


irvinesunday:
Re-opening this [issue](https://github.com/microsoft/OpenAPI.NET.OData/issues/122) here to track fixing paths of functions that have structured or collection-valued parameters.
The right structure should be as described in this [comment](https://github.com/microsoft/OpenAPI.NET.OData/issues/122#issuecomment-968958481).

This [assertion](http://docs.oasis-open.org/odata/odata-openapi/v1.0/cnprd01/odata-openapi-v1.0-cnprd01.html#_Toc472347995) in the ref. doc. is incorrect: 
![image](https://user-images.githubusercontent.com/40403681/160133448-93a61138-7a0a-49f9-8b62-3416fdec6559.png)

The correct format should be to use `{...}` around the parameter name, e.g., `https://graph.microsoft.com/deviceManagement/microsoft.graph.getRoleScopeTagsByIds(ids={ids})` and the `in` property value of the parameter object should be `in: path`.

PR #203, opened 2022-03-25
Fixes paths for functions with structured or collection-valued parameters

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/202

This PR: 
- Replaces parameter alias for structured and collection-valued parameters with `{...}`. 
- Replaces the `in` property value of the parameter object from `in:query` to `in: path`.
- Updates the tests appropriately.

Addresses: https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/913 and https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/902

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
@@ -139,9 +139,7 @@ namespace Microsoft.OpenApi.OData.Edm
                 functionName.Append(function.FullName());
             }
             functionName.Append("(");
-
-            // Structured or collection-valued parameters are represented as a parameter alias in the path template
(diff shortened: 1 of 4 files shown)


baywet (dismissed):

.github/workflows/codeql-analysis.yml
```diff
@@ -19,6 +19,21 @@ on:
...
+        default: 'warning' 
+        type: choice
+        options:
+        - info
+        - warning
+        - debug 
+      tags:
+        description: 'Test scenario tags'
+        required: false 
+        type: boolean
```
baywet, line 36:
```suggestion
  workflow_dispatch:
```

I don't think we need the tagging as all and the log level doesn't seem to be referenced anywhere.

baywet (changes requested):

baywet (approved):

Issue #199, opened 2022-03-17
`ODataTypeCast` paths are not being generated


irvinesunday:
Given this navigation property which has `DerivedTypeContraint` annotations:

![image](https://user-images.githubusercontent.com/40403681/158857606-e4ba9afc-2e3b-4492-a099-624e3c66327d.png)

We should expect to see the below paths, as per the Microsoft Graph V1.0 reference [doc](https://docs.microsoft.com/en-us/graph/api/directory-deleteditems-list?view=graph-rest-1.0&tabs=http#http-request).

![image](https://user-images.githubusercontent.com/40403681/158858061-d34aeab6-a742-4f8b-abdf-425b6ea8db02.png)

Currently, the above paths are missing.

PR #212, opened 2022-04-20
Fixes generation of `ODataTypeCast` paths

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/199

This PR:
- Fixes how OData type cast paths from navigation properties are being generated. The entity type targeting the navigation property is used in the argument in the method `CreateTypeCastPaths()` instead of the entity type declaring the navigation property.
- Updated tests and integration test file due to the above fix.
- Added a new setting `ExpandDerivedTypesNavigationProperties` that can be used to evaluate whether or not to expand derived types navigation properties. When this setting is set to `true`, we end up generating over 219K paths for the beta version of Graph (with a maximum of 1 type cast segment allowed in a path). This is because we have deeply nested derived types in our metadata. This needs to be set to `false` when generating OpenAPI documents for Graph. Trying to limit the  number of type cast segments in a path does not help much. As described above, limiting the number of type cast segments in a path to 1 leads to a path count of over 219K (using the Graph beta integration test file in the test project). All this while ensuring circular generation of type cast segments is avoided and only ensuring unique type cast segments in the path.
- Refactored code in the `Utils.cs` class.

### DIFF
--- a/test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
+++ b/test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
@@ -102,8 +102,8 @@ namespace Microsoft.OpenApi.OData.Edm.Tests
         [InlineData(true, false, false, 7)]
         [InlineData(false, true, false, 5)]
         [InlineData(false, true, true, 4)]
-        [InlineData(true, true, true, 5)]
-        [InlineData(true, true, false, 5)]
+        [InlineData(true, true, true, 8)]
+        [InlineData(true, true, false, 8)]
         public void GetOperationPathsForModelWithDerivedTypesConstraint(bool addAnnotation, bool getNavPropModel, bool requireConstraint, int expectedCount)
         {
             // Arrange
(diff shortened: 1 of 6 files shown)


test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
```diff
@@ -40,15 +40,18 @@ public void GetPathsForGraphBetaModelReturnsAllPaths()
...
+            };
             ODataPathProvider provider = new ODataPathProvider();
 
             // Act
             var paths = provider.GetPaths(model, settings);
 
             // Assert
             Assert.NotNull(paths);
-            Assert.Equal(12261, paths.Count());
+            Assert.Equal(14624, paths.Count());
```
xuzhg, line 54: since you disable the setting, why do you get more paths?

xuzhg:
Others look good to me

test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
irvinesunday: The extra number of paths are because we are now generating OData type cast paths for collection-valued navigation properties that were initially missing, e.g.:

image

What I've turned off in the setting by default is the expansion of OData type cast paths into their navigation properties, for single-valued nav. props. Expanding into those end up adding an absurd number of paths.

baywet (dismissed):

baywet (approved):

Issue #192, opened 2022-03-02
Add support for Graph AGS extensions


adhiambovivian:
Add capability to parse and surface AGS extensions that will help with accessing & exposing hidden properties on AGS such as 'isHidden' which are mainly captured on the CSDL today.

irvinesunday:
Discussion needed as a previous attempt to implement reading custom attributes was debated inconclusively. See the discussion on this PR https://github.com/microsoft/OpenAPI.NET.OData/pull/152

darrelmiller:
The previous attempt to use custom attributes was disbanded because we identified an equivalent OData annotation that that did what we needed.

In this case we have a set of annotations that there are no equivalent OData annotations and as they are implementation specific, there is no design to add standardized capabilities.

In order to support these annotations I propose adding a config setting that accepts a dictionary that maps a qualified XML attribute name to a corresponding OpenAPI x- extension.  When converting the document from CSDL to OpenAPI, we should check on each element if any of the present direct annotations has a corresponding mapping to an OpenAPI x- extension.  If it does, then the extension should be output.  If there are no mappings provided then no extensions will be created.

This will make the feature completely opt-in and only the annotations that are of interest should be translated.  Xml Attributes can only have primitive values, so they should map simply to OpenAPI primitive values.

irvinesunday:
In which OpenAPI object will these extensions be inserted into? `OpenAPIOperation.Extensions`?

darrelmiller:
We are going to have to try and attach the annotation to the closed natural equivalent in OpenAPI.

An attribute (e.g. isHidden) on a navigation property would probably map to an extension on OpenApiPathItem.
An attribute (e.g. isHidden) on structural property would map to a OpenApiSchema

irvinesunday:
@darrelmiller can an element have more than 1 custom XML attribute?

PR #214, opened 2022-04-25
Adds support for reading custom attributes

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/192

This PR:
- Provides a means of reading custom attribute values from the CSDL through the setting `CustomXMLAttributesMapping`, which a client can use to provide a dictionary mapping of custom attribute names and extension names. The values of these custom attribute names will be extracted from the Edm model from an element if it contains the custom attribute. The values will then be mapped to the custom extension values provided in the dictionary and added to the relevant extensions object in the OpenAPI document.
- Tests have been added (and integration files updated) to validate the above.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -151,7 +151,7 @@ namespace Microsoft.OpenApi.OData.Common
 
             foreach (var item in customXMLAttributesMapping)
             {
-                string attributeName = item.Key.Split(':').Last(); // example, ags:IsHidden --> IsHidden
+                string attributeName = item.Key.Split(':').Last(); // example, 'ags:IsHidden' --> 'IsHidden'
                 string extensionName = item.Value;
                 EdmStringConstant customXMLAttribute = model.DirectValueAnnotationsManager.GetDirectValueAnnotations(element)?
                                 .Where(x => x.Name.Equals(attributeName, StringComparison.OrdinalIgnoreCase))?
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -133,7 +133,32 @@ namespace Microsoft.OpenApi.OData.Common
                 || s is ODataStreamContentSegment || s is ODataStreamPropertySegment)).Select(e => e.Identifier));
 
             return navigationPropertyName == null ? value : $"{value}/{navigationPropertyName}";
-        }                
+        }
+
+        /// <summary>
+        /// Attempts to add the specified key and value to the dictionary.
+        /// </summary>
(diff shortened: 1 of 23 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -129,6 +133,74 @@ internal static string NavigationPropertyPath(this ODataPath path, string naviga
...
+
+            Dictionary<string, string> atrributesValueMap = GetCustomXMLAtrributesValueMapping(context.Model, element, context.Settings.CustomXMLAttributesMapping);
+
+            if (atrributesValueMap?.Any() ?? false)
+            {
+                foreach (var item in atrributesValueMap)
+                {
+                    if (!extensions.ContainsKey(item.Key))
+                    {
+                        extensions.Add(item.Key, new OpenApiString(item.Value));
```
baywet, line 163:
```suggestion
                        extensions.TryAdd(item.Key, new OpenApiString(item.Value));
```
and remove the surrounding if block

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: Unfortunately, `IDictionary<string, IOpenApiExtension>` does not contain a definition for `TryAdd`

irvinesunday: Unfortunately, `IDictionary<string, IOpenApiExtension>` does not contain a definition for `TryAdd`

baywet: It's an extension method https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.collectionextensions.tryadd?view=net-6.0

irvinesunday: This is applicable to the [below versions](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.collectionextensions.tryadd?view=net-6.0#applies-to) of the .NET and .NET Standard:

image

The current project is on [.NET Standard 2.0](https://github.com/microsoft/OpenAPI.NET.OData/blob/429e664a27d8ff074d2866eaa5979b0ac142f07f/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj#L15) which doesn't support that extension method:

image

baywet: right, ideally then we'd define our own extension method, this way we avoid duplication, and the day this lib upgrades the target, it'll be easier to identify the places to replace it.

irvinesunday: Resolved.

baywet (approved):

Issue #153, opened 2021-12-09
[Feature] Support for a range of response status codes


peombwa:
Consider adding support for a range of response status codes as described in [OAS 3](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#patterned-fields-1):
> Any HTTP status code can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. A Reference Object can link to a response that is defined in the OpenAPI Object's components/responses section. This field MUST be enclosed in quotation marks (for example, "200") for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character X. For example, 2XX represents all response codes between [200-299]. Only the following range definitions are allowed: 1XX, 2XX, 3XX, 4XX, and 5XX. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.

e.g., 2XX.

This should ideally be an opt-in feature that should be set via `OpenApiConvertSettings`.

darrelmiller:
Response should look something like below. Error payload should be the same as it currently is.

```yaml
openapi: 3.0.1
info:
  title: Wildcard status codes
  version: 1.0.0
paths:
  /:
    get:
      responses:
        2XX:
          description: success
          content:
            application/json:
              schema: 
                type: object
        4XX:
          description: error
          content:
            application/json:
              schema: 
                $ref: "#/components/schemas/error"
        5XX:
          description: error
          content:
            application/json:
              schema: 
                $ref: "#/components/schemas/error"

components:
  schemas:
    error: {}
            

```

irvinesunday:
@peombwa I believe this issue was fixed with this PR: https://github.com/microsoft/OpenAPI.NET.OData/pull/167

peombwa:
@irvinesunday, this issue relates to success (2xx) status code ranges. We've had challenges with certain APIs returning success status codes that are not defined in the OpenAPI files. e.g., https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/907#issuecomment-947183924.

The fix that was made in [adds the ability to define error responses on error codes instead of default by baywet · Pull Request #167 · microsoft/OpenAPI.NET.OData (github.com)](https://github.com/microsoft/OpenAPI.NET.OData/pull/167) was for error responses and not success responses.

PR #215, opened 2022-04-27
Provides support for using the range of response status code `2XX`

irvinesunday:
Closes https://github.com/microsoft/OpenAPI.NET.OData/issues/153

This PR:
- Adds a new convert setting `UseHTTPStatusCodeClass2XX` that can be used to toggle between using a specific response status code or using a status code range identifier - **2XX**. This setting is set to `false` by default.
- Uses this convert setting to toggle between setting the specific status code or using the status code range identifier.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
@@ -5,6 +5,7 @@
 
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.OData.Generator;
+using System.Collections.Generic;
 
 namespace Microsoft.OpenApi.OData.Common;
 
@@ -20,7 +21,8 @@ public static class OpenApiOperationExtensions
     /// <param name="operation">The operation.</param>
     /// <param name="settings">The settings.</param>
     /// <param name="addNoContent">Whether to add a 204 no content response.</param>
-    public static void AddErrorResponses(this OpenApiOperation operation, OpenApiConvertSettings settings, bool addNoContent = false)
+    /// <param name="schema">Optional: The OpenAPI schema of the response.</param>
+    public static void AddErrorResponses(this OpenApiOperation operation, OpenApiConvertSettings settings, bool addNoContent = false, OpenApiSchema schema = null)
     {
         if (operation == null) {
             throw Error.ArgumentNull(nameof(operation));
@@ -34,10 +36,29 @@ public static class OpenApiOperationExtensions
 			operation.Responses = new();
 		}
 
-        if(addNoContent)
+        if (addNoContent)
 		{
-            var statusCode = settings.UseSuccessStatusCodeRange ? Constants.StatusCodeClass2XX : Constants.StatusCode204;
-            operation.Responses.Add(statusCode, Constants.StatusCode204.GetResponse());
+            if (settings.UseSuccessStatusCodeRange && schema != null)
+            {
+                OpenApiResponse response = new()
+                {
+                    Content = new Dictionary<string, OpenApiMediaType>
+                    {
+                        {
+                            Constants.ApplicationJsonMediaType,
+                            new OpenApiMediaType
+                            {
+                                Schema = schema
+                            }
+                        }
+                    }                       
+                };
+                operation.Responses.Add(Constants.StatusCodeClass2XX, response);
+            }
+            else
+            {
+                operation.Responses.Add(Constants.StatusCode204, Constants.StatusCode204.GetResponse());
+            }
 		}
 
         if(settings.ErrorResponsesAsDefault)
--- a/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
@@ -36,7 +36,8 @@ public static class OpenApiOperationExtensions
 
         if(addNoContent)
 		{
-			operation.Responses.Add(Constants.StatusCode204, Constants.StatusCode204.GetResponse());
+            var statusCode = settings.UseHTTPStatusCodeClass2XX ? Constants.StatusCodeClass2XX : Constants.StatusCode204;
+            operation.Responses.Add(statusCode, Constants.StatusCode204.GetResponse());
 		}
 
         if(settings.ErrorResponsesAsDefault)
--- a/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
@@ -36,7 +36,7 @@ public static class OpenApiOperationExtensions
 
         if(addNoContent)
 		{
-            var statusCode = settings.UseHTTPStatusCodeClass2XX ? Constants.StatusCodeClass2XX : Constants.StatusCode204;
+            var statusCode = settings.UseSuccessStatusCodeRange ? Constants.StatusCodeClass2XX : Constants.StatusCode204;
             operation.Responses.Add(statusCode, Constants.StatusCode204.GetResponse());
 		}
 
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
@@ -143,7 +143,8 @@ namespace Microsoft.OpenApi.OData.Generator
             
             if (operation.IsAction() && operation.ReturnType == null)
             {
-                responses.Add(Constants.StatusCode204, Constants.StatusCode204.GetResponse());
+                responses.Add(context.Settings.UseHTTPStatusCodeClass2XX ? Constants.StatusCodeClass2XX : Constants.StatusCode204,
+                    Constants.StatusCode204.GetResponse());
             }
             else
             {
@@ -205,7 +206,7 @@ namespace Microsoft.OpenApi.OData.Generator
                         }
                     }
                 };
-                responses.Add(Constants.StatusCode200, response);
+                responses.Add(context.Settings.UseHTTPStatusCodeClass2XX ? Constants.StatusCodeClass2XX : Constants.StatusCode200, response);
             }
 
             // Both action & function have the default response.
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
@@ -208,8 +208,15 @@ namespace Microsoft.OpenApi.OData.Generator
                 responses.Add(context.Settings.UseSuccessStatusCodeRange ? Constants.StatusCodeClass2XX : Constants.StatusCode200, response);
             }
 
-            // Both action & function have the default response.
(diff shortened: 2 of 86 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
```diff
@@ -36,7 +36,8 @@ public static void AddErrorResponses(this OpenApiOperation operation, OpenApiCon
 
         if(addNoContent)
 		{
-			operation.Responses.Add(Constants.StatusCode204, Constants.StatusCode204.GetResponse());
+            var statusCode = settings.UseHTTPStatusCodeClass2XX ? Constants.StatusCodeClass2XX : Constants.StatusCode204;
+            operation.Responses.Add(statusCode, Constants.StatusCode204.GetResponse());
```
peombwa, line 40: Won't this result in an empty (no content) payload for success status code range - 2xx? i.e., 
``` yml
responses:
  '204':
    description: Success
  default:
```

Clients will handle each success status code accordingly as discussed in https://github.com/Azure/autorest.powershell/issues/853#issuecomment-948617364.

We should always return schema in the response when `UseHTTPStatusCodeClass2XX` is `true`.
``` yml
responses:
  2XX:
    description: success
    content:
      application/json:
        schema: 
          type: object
```

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -237,6 +237,12 @@ public string PathPrefix
         /// </summary>
         public bool RequireRestrictionAnnotationsToGenerateComplexPropertyPaths { get; set; } = true;
 
+        /// <summary>
+        /// Gets/Sets a value indicating whether or not to use the HTTP status code class of 2XX
+        /// to represent all response codes between 200 - 299.
+        /// </summary>
+        public bool UseHTTPStatusCodeClass2XX { get; set; } = false;
```
peombwa, line 244: Consider naming this to something like `UseSuccessStatusCodeRange` to make it implementation agnostic.

src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
irvinesunday: We still do:

image
_with_ `UseHTTPStatusCodeClass2XX = true`


image
_with_ `UseHTTPStatusCodeClass2XX = false`


Nothing in the initial response generation has been modified, just the response status code only (if `UseHTTPStatusCodeClass2XX = true`), everything else remains the same.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
```diff
@@ -143,7 +143,8 @@ public static OpenApiResponses CreateResponses(this ODataContext context, IEdmOp
             
             if (operation.IsAction() && operation.ReturnType == null)
             {
-                responses.Add(Constants.StatusCode204, Constants.StatusCode204.GetResponse());
+                responses.Add(context.Settings.UseHTTPStatusCodeClass2XX ? Constants.StatusCodeClass2XX : Constants.StatusCode204,
```
baywet, line 146: I don't think we should do that for 204. We're effectively loosing precision by doing that going from "no content" to "maybe there's content but it's not described"

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
peombwa: The assumption is that a client will handle success status codes as described in https://github.com/Azure/autorest.powershell/issues/853#issuecomment-948617364. The precision of the status codes is currently lacking, hence the 2xx. e.g., The OpenAPI file describes the response of [updating a user's event](https://docs.microsoft.com/en-us/graph/api/event-update?view=graph-rest-1.0&tabs=http#response-1) as a 204, yet the service returns 200:
image

baywet: sure, my point being, asides from inconsistencies between the service and the OData conventions, if we know by convention some API returns a 204, we should not translate it to a 204.

peombwa: The [OpenAPI spec](https://spec.openapis.org/oas/latest.html#patterned-fields-0) allows for 2XX status codes, covering 200-299, 204 inclusive. If we only describe 204 (no content), clients won't know the schema to deserialize when they receive 200. We can't have 2xx and 204 in the same response object as this will effectively mean 204 is described twice.

The 2XX will be an opt-in feature with the assumption that the client will handle all success status codes, 204 included. This will be similar to how .NET's [IsSuccessStatusCode](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpresponsemessage.issuccessstatuscode?view=net-6.0) is used.

baywet: > If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.

peombwa: I totally missed that part. Yes, in this case, the implementation will need to be updated to include both `204` and `2XX` when we can infer by convention that an API returns 204.

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
irvinesunday: Aah... I see. I will implement your suggestion.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
irvinesunday: Let's recap what needs to be updated in the implementation:
1. When `UseSuccessStatusCodeRange` is set to `true`, for every `204` response we want to add another `2XX` response object with the following format?
```
'204':
    description: Success
2XX:
    description: Success
    content:
      application/json:
        schema: 
          type: object
```
Or do we also want to include a reference to the object in the schema?

2. Does the above apply even to Delete operations with a `204` response status code?

irvinesunday: According to @peombwa's [comment](https://github.com/microsoft/OpenAPI.NET.OData/pull/215#discussion_r862969481) above, sometimes the service does not return the expected `204`.

baywet: wouldn't that mean they are not following OData conventions and that's a fix that needs to happen in the service?

irvinesunday: I agree. Let's hear @peombwa's thoughts on it as well.

peombwa: I beg to differ. Then when should 2xx be used if not for situations like this? If a customer wants a precise response code, they should use `UseSuccessStatusCodeRange = false` (default). There is no requirement for APIs to return 204 for PATCH. The [HTTP spec](https://datatracker.ietf.org/doc/html/rfc5789#section-2.1) says a PATCH method response code can be any success response code (2xx); not just 204:
> The 204 response code is used because the response does not carry a
   message body (which a response with the 200 code would have).  Note
   that other success codes could be used as well.

The [OData spec](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) also allows for 200 or 204 as response codes when updating an entity:
> Upon successful completion the service responds with either [200 OK](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode200OK) and a representation of the updated entity, or [204 No Content](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ResponseCode204NoContent). The client may request that the response SHOULD include a body by specifying a [Prefer header](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_Preferencereturnrepresentationandret) with a value of [return=representation](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_Preferencereturnrepresentationandret), or by specifying the system query options [$select](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_SystemQueryOptionselect) or [$expand](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_SystemQueryOptionexpand).

peombwa: I would say we should:
1. When an operation is an action or function and the return type is null, the response code will only be 204.
2. All other operations should return 2XX for success when `UseSuccessStatusCodeRange = true`.

darrelmiller: FYI, there is nothing wrong with having a specific 200 class responses like 204 and 2XX.  2XX becomes the fallback.

baywet: but from a generation perspective then what would you expect to happen? only one method being generated for the 204 and the 2XX to be ignored? two methods as we assume 2XX also represents a 200/201/202 and we in fact have both 2XX and 204?

darrelmiller: @peombwa @irvinesunday Yes, I think we are now all in agreement.  For actions and functions that return void we should only emit the 204 status code in the responses object.

baywet (changes requested):
Thanks for making the changes. A few minor things left.
csproj changelog entry + PR is conflicting

src/Microsoft.OpenApi.OData.Reader/Common/OpenApiOperationExtensions.cs
irvinesunday: I have added `Content` objects to the responses of POST, PATCH, PUT and UPDATE operations.

test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/ComplexPropertyPatchOperationHandlerTests.cs
```diff
@@ -44,7 +48,8 @@ public void CreateComplexPropertyDeleteOperationReturnsCorrectOperationForSingle
 
 		Assert.NotNull(patch.Responses);
 		Assert.Equal(2, patch.Responses.Count);
-		Assert.Equal(new[] { "204", "default" }, patch.Responses.Select(r => r.Key));
+		var statusCode = useHTTPStatusCodeClass2XX ? "2XX" : "204";
+		Assert.Equal(new[] { statusCode, "default" }, patch.Responses.Select(r => r.Key));
```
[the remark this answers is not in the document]
irvinesunday: This has been added as such:

```
if (useHTTPStatusCodeClass2XX)
{
    Assert.Single(operation.Responses[statusCode].Content);
}
else
{
    Assert.Empty(operation.Responses[statusCode].Content);
}
```

peombwa (dismissed):

baywet (approved):

Issue #221, opened 2022-05-19
OData actions are applied to derived types


peombwa:
OData actions bound to base types are being applied to derived types resolved via OData casting. From my testing, calling an action from an OData cast-derived type is not supported by the service e.g.,
![MicrosoftTeams-image](https://user-images.githubusercontent.com/7061532/169399735-42b179da-9058-41ab-81d8-702c1a454959.png)

This seems to occur even with `ExpandDerivedTypesNavigationProperties = false`.

Related to https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1000.

### Assemblies affected
1.0.10

darrelmiller:
@irvinesunday Is there a setting to disable generating actions on derived types that is independent to the `ExpandDerivedTypesNavigationProperties`?

irvinesunday:
> @irvinesunday Is there a setting to disable generating actions on derived types that is independent to the `ExpandDerivedTypesNavigationProperties`?

@darrelmiller, no we don't have such a setting. 
Despite not being applicable in the Graph service, are actions on derived types allowable in OData?

maisarissi:
Hey folks. This is blocking us from releasing PowerShell SDK, and I saw that this issue was added into the 1.0.11 milestone, instead of being prioritized as part of 1.0.10. Is that right?
Please, can you prioritize this? @CarolKigoonya @irvinesunday @darrelmiller

irvinesunday:
> Hey folks. This is blocking us from releasing PowerShell SDK, and I saw that this issue was added into the 1.0.11 milestone, instead of being prioritized as part of 1.0.10 release. Is that right? Please, can you reevaluate this? @CarolKigoonya @irvinesunday @darrelmiller

The fix will go out in 1.0.11-preview1. I've labelled it as a p1

maisarissi:
When is the release date for 1.0.11? Per my understanding, we are getting close to release the 1.0.10 version and update DevX API to use it, but until we get this fixed, the PowerShell SDK can't be generated.
We have added this item as part of our 1.0.10 tracking working on PS side: https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/1265

irvinesunday:
@maisarissi `v1.0.10` was released and DevX API is currently on this version. `v1.0.11-preview1` is the next candidate and is scheduled to go out with the end of milestone 1.0.11; however, I am fast-tracking this so that we can release it by next week.

[...]

PR #226, opened 2022-05-30
Enables configuring appending bound operations on derived types

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/221

This PR:
- Adds a new convert setting `AppendBoundOperationsOnDerivedTypes` that allows configuring whether or not to append bound operations on derived types. This is enabled by default - `true`.
- Updates the relevant tests to validate use of the above setting flag when generating paths for bound operations on derived types.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
@@ -249,9 +249,9 @@ namespace Microsoft.OpenApi.OData
         public Dictionary<string, string> CustomXMLAttributesMapping { get; set; } = new();
 
         /// <summary>
-        /// Gets/sets a value indicating whether or not to append bound operations on derived property paths.
+        /// Gets/sets a value indicating whether or not to append bound operations on derived types.
         /// </summary>
-        public bool AppendBoundOperationsOnDerivedPaths { get; set; } = true;
+        public bool AppendBoundOperationsOnDerivedTypes { get; set; } = true;
 
         internal OpenApiConvertSettings Clone()
         {
@@ -294,7 +294,7 @@ namespace Microsoft.OpenApi.OData
                 RequireRestrictionAnnotationsToGenerateComplexPropertyPaths = this.RequireRestrictionAnnotationsToGenerateComplexPropertyPaths,
                 ExpandDerivedTypesNavigationProperties = this.ExpandDerivedTypesNavigationProperties,
                 CustomXMLAttributesMapping = this.CustomXMLAttributesMapping,
-                AppendBoundOperationsOnDerivedPaths = this.AppendBoundOperationsOnDerivedPaths
+                AppendBoundOperationsOnDerivedTypes = this.AppendBoundOperationsOnDerivedTypes
             };
 
             return newSettings;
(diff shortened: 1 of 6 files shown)


src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -248,6 +248,11 @@ public string PathPrefix
         /// </summary>
         public Dictionary<string, string> CustomXMLAttributesMapping { get; set; } = new();
 
+        /// <summary>
+        /// Gets/sets a value indicating whether or not to append bound operations on derived types.
+        /// </summary>
+        public bool AppendBoundOperationsOnDerivedTypes { get; set; } = true;
```
baywet, line 254: From what I understand this should actually be AppendBoundOperationsOnDerivedTypeCastSegments.

Now, I'm questioning the value of such a setting. The operations were added to the derived type segments when I implemented the type cast segments because I didn't give too many thoughts about that specific aspect.

The only value of adding these operations is to allow the caller to pass (or get as result) a more specialized derived type instead of the more general base type. But in most OOP languages this should already be possible thanks to auto upcasting (e.g. if I have an operation accepting a directoryObject, I should be able to pass it a User).

Besides increasing the number of path items and the size of SDKs I don't see any other value to this, and since it was only added in 1.0.10, I'd argue we remove that behavior all together, where if one wants to call an operation, they must call it on the non-type cast path item. Thoughts?

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
irvinesunday: Yes, the suggested name for the convert setting makes more sense.

From my tests, the particular path mentioned in the [issue](https://github.com/microsoft/OpenAPI.NET.OData/issues/221), `/servicePrincipals/{servicePrincipal-id}/createdObjects/{directoryObject-id}/microsoft.graph.servicePrincipal/microsoft.graph.addKey` gets its bound operation segment path generated due to this [condition](https://github.com/microsoft/OpenAPI.NET.OData/blob/0bfd9b35b8e9443669053aa4ff53f954f08d008f/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L671). So, even removing this [method](https://github.com/microsoft/OpenAPI.NET.OData/blob/0bfd9b35b8e9443669053aa4ff53f954f08d008f/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L807-L871) that explicitly adds the type cast and operation segments to a navigation property path will not resolve the issue highlighted above. There is also the possibility that a client outside MS Graph might have already taken a dependency on this particular version, hence removing the path might be a breaking change for them. Unless this kind of path is entirely not supported by OData. 

@peombwa any thoughts?

baywet: if we want to keep the setting in the rare case of somebody having already taken a dependency on it in the last 3 weeks, then I'd suggest it's false by default.

baywet (approved):

peombwa (approved):

Issue #218, opened 2022-05-16
odata actions do not have error ranges described


baywet:
OData actions (like assign license for instance), do not have error range responses describes after the conversion process when ErrorResponsesAsDefault is false

https://github.com/microsoft/OpenAPI.NET.OData/blob/04422bf67138891b6bc6822357a8686061ba608e/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs#L227

For instance

```diff
'/groups/{group-id}/microsoft.graph.assignLicense':
    description: Provides operations to call the assignLicense method.
    post:
      tags:
        - groups.Actions
      summary: Invoke action assignLicense
      operationId: groups.group.assignLicense
      requestBody:
        description: Action parameters
        content:
          application/json:
            schema:
              type: object
              properties:
                addLicenses:
                  type: array
                  items:
                    $ref: '#/components/schemas/microsoft.graph.assignedLicense'
                removeLicenses:
                  type: array
                  items:
                    pattern: '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
                    type: string
                    format: uuid
        required: true
      responses:
        '200':
          description: Success
          content:
            application/json:
              schema:
                anyOf:
                  - $ref: '#/components/schemas/microsoft.graph.group'
                nullable: true
-        default:
-          $ref: '#/components/responses/error'
+        4XX:
+          $ref: '#/components/responses/error'
+        5XX:
+          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: action
    parameters:
      - name: group-id
        in: path
        description: 'key: id of group'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: group
```

PR #229, opened 2022-06-02
Fix error ranges for OData actions

millicentachieng:
Fixes (#218)

Add error ranges for OData actions if ErrorResponsesAsDefault is set to false.

image


baywet (dismissed):
thanks for fixing that and sorry for missing it at the first place. Can you add an entry in the csproj changelog please?

peombwa (dismissed):

danielmbaluka (approved):

peombwa (approved):

irvinesunday (approved):

PR #237, opened 2022-06-22
Fix/request body and response representation for ref operations

### LINKED ISSUE #228: Inacurrate request body and response representation for ref POST operations
Thanks to updates in the conversion process we now produce the follow for referenceable collections:

```yaml
  '/groups/{group-id}/members/$ref':
  # omitted
    post:
      tags:
        - groups.directoryObject
      summary: Create new navigation property ref to members for groups
      operationId: groups.CreateRefMembers
      requestBody:
        description: New navigation property ref value
        content:
          application/json:
            schema:
              type: object
              additionalProperties:
                type: object
        required: true
      responses:
        '201':
          description: Created navigation property link.
          content:
            application/json:
              schema:
                type: object
        4XX:
        5XX:
    parameters:
      - name: group-id
        in: path
        description: 'key: id of group'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: group
```

While this works it has three issues:
- It doesn't contain a definition for an odata id property
- The request body is duplicated for every ref POST operation
- The return code should be 204, not 201

We should instead have the following

```diff
  '/groups/{group-id}/members/$ref':
  # omitted
    post:
      tags:
        - groups.directoryObject
      summary: Create new navigation property ref to members for groups
      operationId: groups.CreateRefMembers
      requestBody:
+       $ref: '#/components/requestBodies/refBody'
-        description: New navigation property ref value
-        content:
-          application/json:
-            schema:
-              type: object
-              additionalProperties:
-                type: object
-        required: true
      responses:
+        '204':
-        '201':
-          description: Created navigation property link.
-          content:
-            application/json:
-              schema:
-                type: object
        4XX:
          $ref: '#/components/responses/error'
        5XX:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
    parameters:
      - name: group-id
        in: path
        description: 'key: id of group'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: group
+components:
+  requestBodies
+    refBody:
+      description: New navigation property ref value
+      required: true
+      content:
+        application/json
+          schema
+            $ref: '#/components/schemas/ref'
+  schemas:
+    refBody
+      type: object
+      properties:
+        '@odata.id':
+          type: string
+      additionalProperties:
+        type: object
```

millicentachieng:
Fixes #228

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
@@ -109,5 +109,10 @@ namespace Microsoft.OpenApi.OData.Common
         /// Name used for reference update.
         /// </summary>
         public static string ReferenceUpdateSchemaName = "ReferenceUpdateSchema";
+
+        /// <summary>
+        /// Name use for reference request body
+        /// </summary>
+        public static string ReferenceRequestBodyName = "refBody";
     }
 }
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/RefPostOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/RefPostOperationHandler.cs
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------
+// ------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All rights reserved.
 //  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // ------------------------------------------------------------
@@ -65,29 +65,11 @@ namespace Microsoft.OpenApi.OData.Operation
         /// <inheritdoc/>
         protected override void SetResponses(OpenApiOperation operation)
         {
-            OpenApiSchema schema = new OpenApiSchema
-            {
-                Type = "object"
-            };
-
             operation.Responses = new OpenApiResponses
             {
                 {
-                    Context.Settings.UseSuccessStatusCodeRange ? Constants.StatusCodeClass2XX : Constants.StatusCode201,
-                    new OpenApiResponse
-                    {
-                        Description = "Created navigation property link.",
-                        Content = new Dictionary<string, OpenApiMediaType>
-                        {
-                            {
-                                Constants.ApplicationJsonMediaType,
-                                new OpenApiMediaType
-                                {
-                                    Schema = schema
-                                }
-                            }
-                        }
-                    }
+                    Context.Settings.UseSuccessStatusCodeRange ? Constants.StatusCodeClass2XX : Constants.StatusCode204,
+                    new OpenApiResponse { Description = "Success" } 
                 }
             };
 
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/RefPostOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/RefPostOperationHandler.cs
@@ -1,10 +1,11 @@
-// ------------------------------------------------------------
+﻿// ------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All rights reserved.
 //  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // ------------------------------------------------------------
 
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.OData.Edm;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.OData.Common;
 using Microsoft.OpenApi.OData.Generator;
@@ -38,24 +39,13 @@ namespace Microsoft.OpenApi.OData.Operation
         /// <inheritdoc/>
         protected override void SetRequestBody(OpenApiOperation operation)
         {
-            OpenApiSchema schema = new OpenApiSchema
-            {
-                Type = "object",
-                AdditionalProperties = new OpenApiSchema { Type = "object" }
-            };
-
             operation.RequestBody = new OpenApiRequestBody
             {
-                Required = true,
-                Description = "New navigation property ref value",
-                Content = new Dictionary<string, OpenApiMediaType>
+                UnresolvedReference = true,
+                Reference = new OpenApiReference
                 {
-                    {
-                        Constants.ApplicationJsonMediaType, new OpenApiMediaType
-                        {
-                            Schema = schema
-                        }
-                    }
+                    Type = ReferenceType.RequestBody,
+                    Id = Constants.ReferenceRequestBodyName
                 }
             };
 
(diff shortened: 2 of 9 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
```diff
@@ -109,5 +109,10 @@ internal static class Constants
         /// Name used for reference update.
         /// </summary>
         public static string ReferenceUpdateSchemaName = "ReferenceUpdateSchema";
+
+        /// <summary>
+        /// Name used for reference request body
+        /// </summary>
+        public static string ReferenceRequestBodyName = "refBody";
```
peombwa, line 116: Should we be worried of potential name conflict with workloads entities in the future? Thoughts on having this as `ODataRef` instead?

peombwa: NVM, workload entities will have namespaces in their names. e.g., `microsoft.graph.user`.

peombwa (approved):

irvinesunday:
I think we can also do the same for the `RequestBody` for ref PUT operations?
cc: @baywet

baywet:
I know that patch uses the odata bind style on the parent entity. I'm not sure whether put uses odata bind or is straight on the collection like post. Can you recall?

irvinesunday:
> I think we can also do the same for the `RequestBody` for ref PUT operations? cc: @baywet

Nevermind. I see we already do it with PUT schemas:
https://github.com/microsoft/OpenAPI.NET.OData/blob/35a36642f3f0b95c50e05efe3d6eb4bc0f27270d/src/Microsoft.OpenApi.OData.Reader/Operation/RefPutOperationHandler.cs#L94-L105

baywet:
@irvinesunday correct, thanks for catching that, they are essentially the same since the odata type is nullable. To be specific, inserting a new reference in a collection (e.g. a new group member) via POST or updating a single reference (e.g. manager) via PUT takes the same payload. (and both reply 204)

millicentachieng:
@baywet PUT and POST reference different schemas.
image

src/Microsoft.OpenApi.OData.Reader/Operation/RefPostOperationHandler.cs
```diff
@@ -65,29 +55,11 @@ protected override void SetRequestBody(OpenApiOperation operation)
...
-                            {
-                                Constants.ApplicationJsonMediaType,
-                                new OpenApiMediaType
-                                {
-                                    Schema = schema
-                                }
-                            }
-                        }
-                    }
+                    Context.Settings.UseSuccessStatusCodeRange ? Constants.StatusCodeClass2XX : Constants.StatusCode204,
```
baywet, line 61:
```suggestion
                    Constants.StatusCode204,
```

Since we know by convention it's always going to be a 204, I'd suggest like other places we only ever produce a 204 status code to be the most specific we can.

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Multiple.Schema.OpenApi.V2.json
```diff
@@ -928,12 +928,7 @@
         ],
         "summary": "Create new navigation property ref to Revisions for Documents",
         "operationId": "Documents.CreateRefRevisions",
-        "consumes": [
```
baywet, line 931: I'm a bit confused here, shouldn't it still consume JSON?

baywet (changes requested):
Thanks for making the changes. Re-reading Irvine's comment I realized he was advocating for a naming convention, not reusing the same schema for PUT and POST. A couple of minor things left.

src/Microsoft.OpenApi.OData.Reader/Operation/RefPostOperationHandler.cs
baywet: The specification for reference
http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ChangetheReferenceinaSingleValuedNav
http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_AddaReferencetoaCollectionValuedNavi

src/Microsoft.OpenApi.OData.Reader/Operation/RefPutOperationHandler.cs
```diff
@@ -59,7 +54,7 @@ protected override void SetRequestBody(OpenApiOperation operation)
         /// <inheritdoc/>
         protected override void SetResponses(OpenApiOperation operation)
         {
-    		operation.AddErrorResponses(Context.Settings, true, GetOpenApiSchema());
+    		operation.AddErrorResponses(Context.Settings, true, GetRefUpdateSchema());
```
baywet, line 57: this should also only return 204, align with post

src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
```diff
@@ -110,19 +110,30 @@ internal static class Constants
         /// </summary>
         public static string ReferenceUpdateSchemaName = "ReferenceUpdateSchema";
 
+
+        /// <summary>
+        /// Name used for reference update.
+        /// </summary>
+        public static string ReferencePostSchemaName = "ReferencePostSchema";
```
baywet, line 117: I'd use ReferenceUpdate and ReferenceCreate (or add) without the Schema suffix.
In Kiota this name is going to be used as the type name, and having schema in it makes for an ugly class name. Thoughts?

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
millicentachieng: We've had `ReferenceUpdateSchema` so I was following the same convention. So maybe `ReferenceCreateSchema`? I do not have a problem with appending the word schema. It provides clarity and prevents ambiguity.

baywet: In the OpenApi Description yes, but in the code people will have to do something like

```CSharp

var myVal = new ReferenceCreateSchema();
```

Which feels odd, don't you think? Especially since it'd be so easy for us to make it.

```CSharp

var myVal = new ReferenceCreate();
```

millicentachieng: I see. Updating...

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/Multiple.Schema.OpenApi.V2.json
millicentachieng: It should consume JSON. The v2 deserializer is not properly handling ref requestBodies. I've created this [issue](https://github.com/microsoft/OpenAPI.NET/issues/932) to track the problem.

baywet: thanks for logging the issue, so this is "not this lib's problem" at this point. resolving.

baywet (approved):

irvinesunday (approved):

peombwa (approved):

PR #244, opened 2022-07-04
Update CODEOWNERS

irvinesunday:
Add @baywet and @millicentachieng as codeowners


baywet (approved):
Also changed some permissions to favor using teams vs one off permissions allocations. This fixed @zengin being reported as error in here (he didn't have permissions on the repo)

Issue #243, opened 2022-06-28
Discriminator object refers to missing property


andrueastman:
Latest update of the conversion library includes adding some discriminator mappings to the openApi document. Possibly related to changes fixing https://github.com/microsoft/OpenAPI.NET.OData/issues/233

Taking a look at the previous description of the `apiAuthenticationConfigurationBase`

```yaml
    microsoft.graph.apiAuthenticationConfigurationBase:
      title: apiAuthenticationConfigurationBase
      type: object
```

And the new description 

``` yaml
    microsoft.graph.apiAuthenticationConfigurationBase:
      title: apiAuthenticationConfigurationBase
      type: object
      discriminator:
        propertyName: '@odata.type'
        mapping:
          '#microsoft.graph.basicAuthentication': '#/components/schemas/microsoft.graph.basicAuthentication'
          '#microsoft.graph.clientCertificateAuthentication': '#/components/schemas/microsoft.graph.clientCertificateAuthentication'
          '#microsoft.graph.pkcs12Certificate': '#/components/schemas/microsoft.graph.pkcs12Certificate'
```

The discriminator mapping mentions the propertyName `@odata.type` but the property collection (this instance has no properties) does not contain the property `@odata.type`. 

This leads to validation errors as described in https://github.com/microsoft/kiota/issues/1668. 

The conversion library should probably make sure the property exists before adding the mapping.

andrueastman:
According to https://spec.openapis.org/oas/v3.1.0#fixed-fields-20,

> The discriminator object is legal only when using one of the composite keywords oneOf, anyOf, allOf.

baywet:
[related](https://github.com/microsoft/OpenAPI.NET/issues/402#issuecomment-1168670671)

The conversion mechanism should add the `@odata.type` field to the required field list in the type the field is defined.

Also related #240.


Here the spec contradicts itself by providing the following example where Pet does not have any/all/one of but has a discriminator node. I do think we should ignore this directive from the spec as if you follow it, the only way to implement inheritance + discriminator + mapping would be to have the discriminator field defined on each derived type AND have the mapping entry containing the derived type it's defined on, which would be both redundant, and make for a complex guessing game when looking at the base type. What do you think @darrelmiller ?

```yaml
components:
  schemas:
    Pet:
      type: object
      required:
      - petType
      properties:
        petType:
          type: string
      discriminator:
        propertyName: petType
        mapping:
          dog: Dog
    Cat:
      allOf:
[...]

PR #246, opened 2022-07-04
Adds `@odata.type` property and makes this a required property in schemas that have discriminator objects

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/243

This PR does the following:

Given an example of this schema, which is a base type, which exposes the discriminator mappings as shown:

MicrosoftTeams-image (2)

The following is accomplished:

1. Adds the @odata.type property in the properties' object
2. Marks the @odata.type property as required
3. In the @odata.type property, sets the default value to '#microsoft.graph.entity
4. Updates tests to validate the above.

After the above, the schema should appear as below:

image

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
@@ -102,7 +102,7 @@ namespace Microsoft.OpenApi.OData.Generator
                     Properties = new Dictionary<string, OpenApiSchema>
                     {
                         {"@odata.id", new OpenApiSchema { Type = "string", Nullable = false }},
-                        {"@odata.type", new OpenApiSchema { Type = "string", Nullable = true }},
+                        {Constants.OdataType, new OpenApiSchema { Type = "string", Nullable = true }},
                     }
                 };
             }
@@ -435,7 +435,6 @@ namespace Microsoft.OpenApi.OData.Generator
             {
                 // The discriminator object is added to structured types which have derived types.
                 OpenApiDiscriminator discriminator = null;
-                string discriminatorPropertyName = "@odata.type";
                 if (context.Settings.EnableDiscriminatorValue && derivedTypes.Any())
                 {
                     Dictionary<string, string> mapping = derivedTypes
@@ -450,7 +449,7 @@ namespace Microsoft.OpenApi.OData.Generator
 
                     discriminator = new OpenApiDiscriminator
                     {
-                        PropertyName = discriminatorPropertyName,
+                        PropertyName = Constants.OdataType,
                         Mapping = mapping
                     };
                 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -435,6 +435,7 @@ internal static OpenApiSchema CreateSchemaTypeSchema(this ODataContext context,
             {
                 // The discriminator object is added to structured types which have derived types.
                 OpenApiDiscriminator discriminator = null;
+                string discriminatorPropertyName = "@odata.type";
```
baywet, line 438: move to a const outside of the method.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -473,6 +474,16 @@ internal static OpenApiSchema CreateSchemaTypeSchema(this ODataContext context,
                     AnyOf = null
                 };
 
+                if (schema.Discriminator != null)
+                {
+                    schema.Properties.TryAdd(discriminatorPropertyName, new OpenApiSchema()
+                    {
+                        Type = "string"                        
+                    });
+                    schema.Properties[discriminatorPropertyName].Default = new OpenApiString("#" + structuredType.FullTypeName());
```
baywet, line 483:
```suggestion
                        Type = "string",
                       .Default = new OpenApiString("#" + structuredType.FullTypeName()),
                    });
```

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: No. I intentionally set the `Default` outside the `TryAdd()` method so as to capture for all cases, including those where the `discriminatorPropertyName` -> `@odata.type` property already exists in the schema.

baywet: in what scenario could the property could have already been added?

irvinesunday: In your [comment](https://github.com/microsoft/OpenAPI.NET.OData/issues/243#issuecomment-1168726146) in the issue you mentioned that some base types are missing this property and that we should walk the inheritance tree to make sure the property is present and add it if not.

baywet: Ah my bad, I should probably have added another comment with the latest requirement to clear any confusion.
The current implementation adds the property to all entity or complex types (which is what we now want). TryAdd would never have checked the presence of the property in a base schema, for that we'd have had to walk the allOf properties recursively. But since we don't want to add the property only on the base type anymore this requirement is moot.
I can't foresee a case where the odata type property would already be here besides it being described in the CSDL, and using convention reserved property names would be horribly wrong, and the conversion should fail.

Bottom line, move the default above, and if the tryadd returns false, throw an invalid operation exception saying the property is already present, which is not expected, and the CSDL should be verified.

I hope this makes things clearer, don't hesitate if you have follow up questions.

baywet (approved):

PR #247, opened 2022-07-06
Represent nullable references within anyOf correctly

### LINKED ISSUE #190: Nullable references are not represented correctly
Currently nullable references are presented like this:

```yaml
             schema:
                anyOf:
                  - $ref: '#/components/schemas/microsoft.graph.domain'
                nullable: true
```

The OpenAPI specification describes the impact of nullable as:
> A true value adds "null" to the allowed type specified by the type keyword, only if type is explicitly defined within the same Schema Object. Other Schema Object constraints retain their defined behavior, and therefore may disallow the use of null as a value. A false value leaves the specified or default type unmodified. The default value is false.

Based on this definition, to allow this schema to be valid if either the referenced schema is true, or a null object is true, then the OpenAPI should look like this:
 
```yaml
             schema:
                anyOf:
                  - $ref: '#/components/schemas/microsoft.graph.domain'
                  - type: object
                    nullable: true
```

millicentachieng:
Fixes #190

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
@@ -420,7 +420,6 @@ namespace Microsoft.OpenApi.OData.Generator
             Debug.Assert(typeReference != null);
 
             OpenApiSchema schema = new OpenApiSchema();
-            schema.Nullable = typeReference.IsNullable;
             schema.Reference = null;
 
             if (context.Settings.OpenApiSpecVersion >= OpenApiSpecVersion.OpenApi3_0)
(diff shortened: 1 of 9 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
```diff
@@ -420,7 +420,6 @@ private static OpenApiSchema CreateEnumTypeSchema(this ODataContext context, IEd
             Debug.Assert(typeReference != null);
 
             OpenApiSchema schema = new OpenApiSchema();
-            schema.Nullable = typeReference.IsNullable;
             schema.Reference = null;
 
             if (context.Settings.OpenApiSpecVersion >= OpenApiSpecVersion.OpenApi3_0)
```
millicentachieng, line 425: I'm not sure why the `typeReference.IsNullable` check wasn't done here as in one other place. Why did we have AnyOf when there's only one item in the list?

baywet: Agreed, I'd argue we should have an any of only if nullable is true, otherwise we're just bloating the description.

baywet (commented):
holding until we get an answer to this comment https://github.com/microsoft/OpenAPI.NET.OData/issues/190#issuecomment-1177539141

irvinesunday (approved):

baywet (approved):

PR #255, opened 2022-07-16
Fix response schemas of actions and functions that return a collection to contain the nextLink property.

### LINKED ISSUE #231: Response schemas of actions and functions that return a collection do not contain the nextLink property.
Related to https://github.com/microsoft/OpenAPI.NET.OData/issues/116

### Assemblies affected

1.0.10

### Steps to reproduce

https://github.com/microsoft/OpenAPI.NET.OData/issues/116 involved fixing the response shape for the collections returned in actions/functions.

However, these collections may be pageable and may contain the `@odata.nextLink` property which are not catered for at the moment. Ideally, this should be part of response schema as is the case for other collection properties if possible.

millicentachieng:
Fixes #231 
Add pageable extension to responses of actions and functions that return a collection

### DIFF
--- a/test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/EdmActionOperationHandlerTests.cs
+++ b/test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/EdmActionOperationHandlerTests.cs
@@ -54,7 +54,7 @@ namespace Microsoft.OpenApi.OData.Operation.Tests
         }
 
         [Fact]
-        public void CreateOperationForEdmActionReturnsCorrectOperationHierarhicalClass()
+        public void CreateOperationForEdmActionReturnsCorrectOperationHierarchicalClass()
         {
             // Arrange
             IEdmModel model = EdmModelHelper.ContractServiceModel;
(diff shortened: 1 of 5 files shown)


test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/EdmActionOperationHandlerTests.cs
```diff
@@ -296,5 +296,44 @@ public void OperationRestrictionsTermWorksToCreateOperationForEdmAction(bool ena
...
+
+        [Theory]
+        [InlineData("getMailTips", true)] // returns collection
+        [InlineData("getMailTips", false)] // returns collection
+        [InlineData("findMeetingTimes", true)] // does not return collection
+        public void CreateOperationForEdmActionWithCollectionReturnTypeContainsXMsPageableExtension(string actionName, bool enablePagination)
+        {
+            // Arrange
+            IEdmModel model = EdmModelHelper.GraphBetaModel;
+            OpenApiConvertSettings settings = new OpenApiConvertSettings
```
irvinesunday, line 308:
```suggestion
            OpenApiConvertSettings settings = new()
```

peombwa (approved):

PR #260, opened 2022-08-03
Add the documentation links from Links annotations in CSDL to OpenAPI Operations

### LINKED ISSUE #230: Add the documentation links from `Links` annotations in CSDL to OpenAPI Operations
Given the `Links` annotation in the CSDL:
image

We would want to add the external doc. link (from the **href** property) to the generated OpenAPI doc. in the `ExternalDocs` property of operations:
image

millicentachieng:
Closes #230 

Example:

```yaml

  /admin/serviceAnnouncement/healthOverviews:
    description: Provides operations to manage the healthOverviews property of the microsoft.graph.serviceAnnouncement entity.
    get:
      tags:
        - admin.serviceAnnouncement.serviceHealth
      summary: Get healthOverviews from admin
      description: 'A collection of service health information for tenant. This property is a contained navigation property, it is nullable and readonly.'
      externalDocs:
        description: Find more info here
        url: https://docs.microsoft.com/graph/api/serviceannouncement-list-healthoverviews?view=graph-rest-1.0
      operationId: admin.serviceAnnouncement.ListHealthOverviews
      parameters:
         #truncated
      responses:
        '200':
          $ref: '#/components/responses/microsoft.graph.serviceHealthCollectionResponse'
        4XX:
        5XX:
```

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
@@ -272,23 +272,14 @@ namespace Microsoft.OpenApi.OData.Operation
         {
             if (Context.Settings.CustomHttpMethodLinkRelMapping != null)
             {
-                LinkRelKey? key = null;
-                if (OperationType == OperationType.Get)
+                LinkRelKey? key = OperationType switch
                 {
-                    key = Path.LastSegment?.Kind ==  ODataSegmentKind.Key ? LinkRelKey.ReadByKey : LinkRelKey.List;
-                }
-                else if (OperationType == OperationType.Patch)
-                {
-                    key = LinkRelKey.Update;
-                }
-                else if (OperationType == OperationType.Post)
-                {
-                    key = LinkRelKey.Create;
-                }
-                else if (OperationType == OperationType.Delete)
-                {
-                    key = LinkRelKey.Delete;
-                }
+                    OperationType.Get => Path.LastSegment?.Kind ==  ODataSegmentKind.Key ? LinkRelKey.ReadByKey : LinkRelKey.List,
+                    OperationType.Post => LinkRelKey.Create,
+                    OperationType.Patch => LinkRelKey.Update,
+                    OperationType.Delete => LinkRelKey.Delete,
+                    _ => null,
+                };
 
                 if (key != null)
                 {
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/EntitySetPostOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/EntitySetPostOperationHandler.cs
@@ -11,7 +11,6 @@ using Microsoft.OpenApi.OData.Common;
 using Microsoft.OpenApi.OData.Edm;
 using Microsoft.OpenApi.OData.Generator;
 using Microsoft.OpenApi.OData.Vocabulary.Capabilities;
-using Microsoft.OpenApi.OData.Vocabulary.Core;
 
 namespace Microsoft.OpenApi.OData.Operation
 {
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/MediaEntityOperationalHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/MediaEntityOperationalHandler.cs
@@ -25,8 +25,7 @@ namespace Microsoft.OpenApi.OData.Operation
         /// Gets/Sets the NavigationSource segment
         /// </summary>
         protected ODataNavigationSourceSegment NavigationSourceSegment { get; private set; }
-
-        
+    
         /// <summary>
         /// Gets/Sets flag indicating whether path is navigation property path
         /// </summary>


src/Microsoft.OpenApi.OData.Reader/Edm/EdmAnnotationExtensions.cs
```diff
@@ -256,6 +258,46 @@ public static IEnumerable<T> GetCollection<T>(this IEdmModel model, IEdmVocabula
...
         }
 
+        /// <summary>
+        /// Gets the links record value (a complex type) for the given <see cref="IEdmVocabularyAnnotatable"/>.
+        /// </summary>
+        /// <param name="model">The Edm model.</param>
+        /// <param name="target">The Edm target.</param>
+        /// <param name="operationType">The Operation type.</param>
+        /// <param name="path">The OData path.</param>
+        /// <returns>Null or the </returns>
```
irvinesunday, line 268: Incomplete :)

src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationImportOperationHandler.cs
```diff
@@ -152,5 +153,22 @@ internal static string PathAsString(IEnumerable<string> path)
         {
             return String.Join("/", path);
         }
+
+        /// <inheritdoc/>
+        protected override void SetExternalDocs(OpenApiOperation operation)
+        {
+            if (Context.Settings.ShowExternalDocs)
```
baywet, line 160:
```suggestion
            if (Context.Settings.ShowExternalDocs && Context.Model.GetLinkRecord(EdmOperationImport, OperationType, Path) is Link externalDocs)
```

(and remove the other condition)
(and repeat to other places)

src/Microsoft.OpenApi.OData.Reader/Vocabulary/Core/CoreConstants.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Org.OData.Core.V1.Revisions
+        /// </summary>
+        public const string Revisions = "Org.OData.Core.V1.Revisions";
+
+        /// <summary>
+        /// Link relation types.
+        /// </summary>
+        public static class LinkRel
```
baywet, line 26: instead of having hardcoded constants to Microsoft Graph, I'd have a configuration object on the settings, which default to those values. This way people using it for other APIs can always change the links.

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -179,6 +179,11 @@ public string PathPrefix
         /// </summary>
         public bool ShowMsDosGroupPath { get; set; } = true;
 
+        /// <summary>
+        /// Gets/sets links to external documentation for operations
+        /// </summary>
+        public bool ShowExternalDocs { get; set; } = true;
```
baywet, line 185: this could be a getter that checks the configuration object I was mentioning in the other comment is not null. This way you don't have to update the code all over.

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -259,6 +264,7 @@ public string PathPrefix
         /// </summary>
         public bool UseSuccessStatusCodeRange { get; set; } = false;
 
+
         internal OpenApiConvertSettings Clone()
         {
             var newSettings = new OpenApiConvertSettings
```
baywet, line 270: clone should contain the configuration value

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Edm/EdmAnnotationExtensions.cs
millicentachieng: Thanks for catching that :)

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
millicentachieng: Sure

src/Microsoft.OpenApi.OData.Reader/Vocabulary/Core/CoreConstants.cs
baywet: correct

src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationImportOperationHandler.cs
```diff
@@ -152,5 +153,18 @@ internal static string PathAsString(IEnumerable<string> path)
         {
             return String.Join("/", path);
         }
+
+        /// <inheritdoc/>
+        protected override void SetExternalDocs(OpenApiOperation operation)
+        {
+            if (Context.Settings.ShowExternalDocs && Context.Model.GetLinkRecord(EdmOperationImport, OperationType, Path) is Link externalDocs)
+            {
+                operation.ExternalDocs = operation.ExternalDocs = new OpenApiExternalDocs()
```
baywet, line 162:
```suggestion
                operation.ExternalDocs = new OpenApiExternalDocs()
```

src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
```diff
@@ -198,5 +199,20 @@ protected override void AppendCustomParameters(OpenApiOperation operation)
...
         }
+
+
+
+        /// <inheritdoc/>
+        protected override void SetExternalDocs(OpenApiOperation operation)
+        {
+            if (Context.Settings.ShowExternalDocs && Context.Model.GetLinkRecord(EdmOperation, OperationType, Path) is Link externalDocs)
+            {
+                operation.ExternalDocs = operation.ExternalDocs = new OpenApiExternalDocs()
```
baywet, line 210:
```suggestion
                operation.ExternalDocs = new OpenApiExternalDocs()
```

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
millicentachieng: @baywet Wouldn't that mean this setting will then be read-only?

baywet: yeah and people would enable it by setting the configuration object with all the links, and disable it by setting that configuration object to null. What do you think?

millicentachieng: The scenario I'm thinking about is DevX API where we are generating Open API docs for multiple use cases. Being able to update the setting ShowExternalDocs would be much easier to do than adding or removing the LinkRelMapping setting. Also, it's more intuitive to update ShowExternalDocs than LinkRelMapping if you want to show links to external docs or not.

src/Microsoft.OpenApi.OData.Reader/Operation/OperationHandler.cs
```diff
@@ -254,5 +264,38 @@ protected static void AppendCustomParameters(OpenApiOperation operation, IList<C
...
+
+        /// <summary>
+        /// Set link relation type to be used to get external docs link for path operation
+        /// </summary>
+        protected virtual void SetCustomLinkRelType()
+        {
+            if (Context.Settings.CustomHttpMethodLinkRelMapping != null)
+            {
+                LinkRelKey? key = null;
+                if (OperationType == OperationType.Get)
```
baywet, line 276: nit: you could have used pattern matching ( value switch ...)

baywet (approved):
thanks for taking in the changes

baywet (approved):

Issue #249, opened 2022-07-07
Capability annotations are ignored for certain nav properties


peombwa:
Give a CSDL with the following capability annotations:
``` xml
<Annotations Target="microsoft.graph.user/photo">
    <Annotation Term="Org.OData.Capabilities.V1.ChangeTracking">
        <Record>
        <PropertyValue Property="Supported" Bool="false" />
        </Record>
    </Annotation>
    <Annotation Term="Org.OData.Capabilities.V1.DeleteRestrictions">
        <Record>
        <PropertyValue Property="Deletable" Bool="false" />
        </Record>
    </Annotation>
    <Annotation Term="Org.OData.Capabilities.V1.ExpandRestrictions">
        <Record>
        <PropertyValue Property="Expandable" Bool="false" />
        </Record>
    </Annotation>
    <Annotation Term="Org.OData.Capabilities.V1.InsertRestrictions">
        <Record>
        <PropertyValue Property="Insertable" Bool="false" />
        </Record>
    </Annotation>
    <Annotation Term="Org.OData.Capabilities.V1.SearchRestrictions">
        <Record>
        <PropertyValue Property="Searchable" Bool="false" />
        </Record>
    </Annotation>
</Annotations>
```

### Current Behavior
We generate an OpenAPI document with the following operations:
![image](https://user-images.githubusercontent.com/7061532/177828845-9385ee85-4c3b-4d83-b3e8-8a365ecc80d3.png)

### Expected Behavior
![image](https://user-images.githubusercontent.com/7061532/177830779-dd302abc-1077-4813-95c7-21f74033b1a7.png)

PR #261, opened 2022-08-05
Retrieves `RestrictedProperties` annotations annotated directly on navigations properties

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/249

This PR:

- Retrieves `RestrictedProperties` annotations annotated directly on navigations properties. 
  If the entity type targeted by this navigation property also has `RestrictedProperties` annotations, this will take precedence first.
- Retrieves `IndexableByKey` annotation annotated directly on a navigation property. 
- Updates tests appropriately.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
@@ -403,7 +403,21 @@ namespace Microsoft.OpenApi.OData.Edm
             visitedNavigationProperties.Push(navPropFullyQualifiedName);
 
             // Check whether a collection-valued navigation property should be indexed by key value(s).
-            bool indexableByKey = restriction?.IndexableByKey ?? true;
+            bool indexableByKey = true;
+
+            if (restriction?.IndexableByKey != null)
+            {
+                indexableByKey = (bool)restriction.IndexableByKey;
+            }
+            else
+            {
+                // Find indexability annotation annotated directly via NavigationPropertyRestriction
+                IEdmVocabularyAnnotation indexabilityAnnotation = _model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(navigationProperty, CapabilitiesConstants.IndexableByKey).FirstOrDefault();
+                if (indexabilityAnnotation != null)
+                {
+                    indexableByKey = ((IEdmBooleanValue)indexabilityAnnotation.Value).Value;
+                }
+            }
 
             if (indexableByKey)
             {
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyDeleteOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyDeleteOperationHandler.cs
@@ -21,14 +21,14 @@ namespace Microsoft.OpenApi.OData.Operation
     {
         /// <inheritdoc/>
         public override OperationType OperationType => OperationType.Delete;
+
         private DeleteRestrictionsType _deleteRestriction;
 
         /// <inheritdoc/>
         protected override void Initialize(ODataContext context, ODataPath path)
         {
             base.Initialize(context, path);
-            _deleteRestriction = Restriction?.DeleteRestrictions ??
-                Context.Model.GetRecord<DeleteRestrictionsType>(NavigationProperty, CapabilitiesConstants.DeleteRestrictions);
+            _deleteRestriction = GetRestrictionAnnotation(CapabilitiesConstants.DeleteRestrictions) as DeleteRestrictionsType;
         }
 
         /// <inheritdoc/>
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyDeleteOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyDeleteOperationHandler.cs
@@ -6,7 +6,9 @@
 using System.Linq;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.OData.Common;
+using Microsoft.OpenApi.OData.Edm;
 using Microsoft.OpenApi.OData.Generator;
+using Microsoft.OpenApi.OData.Vocabulary.Capabilities;
 
 namespace Microsoft.OpenApi.OData.Operation
 {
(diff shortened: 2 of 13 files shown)


src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -403,7 +403,21 @@ private void RetrieveMediaEntityStreamPaths(IEdmEntityType entityType, ODataPath
...
+            }
+            else
+            {
+                // Find indexability annotation annotated directly via NavigationPropertyRestriction
+                IEdmVocabularyAnnotation indexabilityAnnotation = _model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(navigationProperty, CapabilitiesConstants.IndexableByKey).FirstOrDefault();
+                if (indexabilityAnnotation != null)
+                {
+                    indexableByKey = ((IEdmBooleanValue)indexabilityAnnotation.Value).Value;
+                }
+            }
```
baywet, line 420:
```suggestion
            else if (_model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(navigationProperty, CapabilitiesConstants.IndexableByKey).FirstOrDefault() is IEdmVocabularyAnnotation indexabilityAnnotation && indexabilityAnnotation.Value is IEdmBooleanValue booleanValue)
            {
                // Find indexability annotation annotated directly via NavigationPropertyRestriction
                indexableByKey = booleanValue.Value;
            }
```

src/Microsoft.OpenApi.OData.Reader/PathItem/NavigationPropertyPathItemHandler.cs
```diff
@@ -80,16 +83,26 @@ protected override void SetOperations(OpenApiPathItem item)
...
                     {
-                        UpdateRestrictionsType updateEntity = Context.Model.GetRecord<UpdateRestrictionsType>(_entityType);
-                        if (updateEntity?.IsUpdatable ?? true)
+                        UpdateRestrictionsType entityUpdateRestriction = Context.Model.GetRecord<UpdateRestrictionsType>(_navPropEntityType);
+                        UpdateRestrictionsType navPropUpdateRestriction = Context.Model.GetRecord<UpdateRestrictionsType>(NavigationProperty);
+                        if ((entityUpdateRestriction?.IsUpdatable ?? true) ||
+                            (navPropUpdateRestriction?.IsUpdatable ?? true))
                         {
-                            AddUpdateOperation(item, restriction);
+                            UpdateRestrictionsType updateRestrictionType = entityUpdateRestriction ?? navPropUpdateRestriction;
```
baywet, line 91:
```suggestion
                            UpdateRestrictionsType updateRestrictionType = navPropUpdateRestriction ?? entityUpdateRestriction;
```

So the nav prop restriction can act as an override to the entity one.

src/Microsoft.OpenApi.OData.Reader/PathItem/NavigationPropertyPathItemHandler.cs
```diff
@@ -80,16 +83,26 @@ protected override void SetOperations(OpenApiPathItem item)
                 {
                     if (LastSegmentIsKeySegment)
                     {
-                        UpdateRestrictionsType updateEntity = Context.Model.GetRecord<UpdateRestrictionsType>(_entityType);
-                        if (updateEntity?.IsUpdatable ?? true)
+                        UpdateRestrictionsType entityUpdateRestriction = Context.Model.GetRecord<UpdateRestrictionsType>(_navPropEntityType);
+                        UpdateRestrictionsType navPropUpdateRestriction = Context.Model.GetRecord<UpdateRestrictionsType>(NavigationProperty);
+                        if ((entityUpdateRestriction?.IsUpdatable ?? true) ||
+                            (navPropUpdateRestriction?.IsUpdatable ?? true))
```
baywet, line 89:
```suggestion
                        if ((entityUpdateRestriction?.IsUpdatable ?? true) &&
                            (navPropUpdateRestriction?.IsUpdatable ?? true))
```

Otherwise, if one even is false, it'll always be added in the absence of the other.

(replicate for the other operations types)

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyDeleteOperationHandler.cs
```diff
@@ -19,14 +21,23 @@ internal class NavigationPropertyDeleteOperationHandler : NavigationPropertyOper
...
         public override OperationType OperationType => OperationType.Delete;
+        private DeleteRestrictionsType _deleteRestriction;
+
+        /// <inheritdoc/>
+        protected override void Initialize(ODataContext context, ODataPath path)
+        {
+            base.Initialize(context, path);
+            _deleteRestriction = Restriction?.DeleteRestrictions ??
+                Context.Model.GetRecord<DeleteRestrictionsType>(NavigationProperty, CapabilitiesConstants.DeleteRestrictions);
+        }
```
xuzhg, line 32: I noticed these codes are repeated in all navigation property handler, could it be better in the base class?

src/Microsoft.OpenApi.OData.Reader/Operation/RefDeleteOperationHandler.cs
```diff
@@ -18,14 +20,23 @@ internal class RefDeleteOperationHandler : NavigationPropertyOperationHandler
...
         public override OperationType OperationType => OperationType.Delete;
+        private DeleteRestrictionsType _deleteRestriction;
+
+        /// <inheritdoc/>
+        protected override void Initialize(ODataContext context, ODataPath path)
+        {
+            base.Initialize(context, path);
+            _deleteRestriction = Restriction?.DeleteRestrictions ??
+                Context.Model.GetRecord<DeleteRestrictionsType>(NavigationProperty, CapabilitiesConstants.DeleteRestrictions);
+        }
```
xuzhg, line 31: same here, could be better put it in the base class.

src/Microsoft.OpenApi.OData.Reader/PathItem/NavigationPropertyPathItemHandler.cs
```diff
@@ -44,7 +44,10 @@ internal class NavigationPropertyPathItemHandler : PathItemHandler
         /// </summary>
         protected bool LastSegmentIsRefSegment { get; private set; }
 
-        private IEdmEntityType _entityType;
+        /// <summary>
```
xuzhg, line 47: insert new line between fields.

xuzhg:
Leave comments, others looks good to me.

src/Microsoft.OpenApi.OData.Reader/PathItem/NavigationPropertyPathItemHandler.cs
irvinesunday: If we have this example:
```
<NavigationProperty Name="recoveryKeys" Type="Collection(graph.bitlockerRecoveryKey)" ContainsTarget="true">
          <Annotation Term="Org.OData.Core.V1.Description" String="The recovery keys associated with the bitlocker entity." />
        </NavigationProperty>
```

Where the navigation property `recoveryKeys` and the entity type targeted by the navigation property `bitlockerRecoveryKey` are both annotated with conflicting restrictions annotations, shouldn't the navigation property's entity type restriction annotation take precedence?

src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyDeleteOperationHandler.cs
irvinesunday: Yes. I have refactored these to be in the base class.

src/Microsoft.OpenApi.OData.Reader/PathItem/NavigationPropertyPathItemHandler.cs
irvinesunday: Resolved

src/Microsoft.OpenApi.OData.Reader/PathItem/NavigationPropertyPathItemHandler.cs
irvinesunday: Do we want navigation restrictions to override the entity type (that's targeted by the navigation property) restrictions? 
I haven't seen any use case yet for this and I am not sure about the precedence.
@darrelmiller ?

darrelmiller: Precedence is nav property overrides entity type.  A scenario would be "this entity can be deleted", however, "you can't delete the entity from this nav property collection".

baywet (approved):

Issue #245, opened 2022-07-04
translation to anyOf and oneOf has been inverted throughout the conversion process


baywet:
According to the [JSON schema spec](https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.5.4)

## AnyOf

> An instance validates successfully against this keyword if it
   validates successfully against *at least one schema* defined by this
   keyword's value.

## OneOf

> An instance validates successfully against this keyword if it
   validates successfully against *exactly one schema* defined by this
   keyword's value.

There are a number of places where the conversion library is using anyOf when it should be using instead oneOf, and it seems the confusion originates from the early days of the library, as [this comment](https://github.com/microsoft/OpenAPI.NET.OData/blob/bdf4241355caba36c7180dc4fd46064a870a161c/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs#L143) would indicate.

If we take the example of OData double to illustrate this issue, the library currently produces this.

```yaml
                zoom:
                  anyOf:
                    - type: number
                    - type: string
                    - enum:
                        - '-INF'
                        - INF
                        - NaN
                  format: double
                  nullable: true
```

Which could be translated to a human sentence to "any instance that is a number (rational) or a string or -INF or INF or NaN is a double".
But the missing details here is that we're using "or" and not "xor" (exclusive or).
Semantically speaking, the value will (and should) never be a number AND -INF for example.

I believe the right *human* sentence should be "any instance that is a number (rational) or a string or -INF or INF or NaN to exactly one of these options is a double". Which would result in the following.

```yaml
                zoom:
                  oneOf:
                    - type: number
                    - type: string
                    - enum:
                        - '-INF'
                        - INF
                        - NaN
                  format: double
                  nullable: true
```

This is even more evident with the geometric data structures where a geometric information can be a point or a line but not both (mathematically speaking, a line is defined a continuous set of points).
https://github.com/microsoft/OpenAPI.NET.OData/blob/bdf4241355caba36c7180dc4fd46064a870a161c/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSpatialTypeSchemaGenerator.cs#L217-L225

# TODO

- [x] update geometric points (link above) to oneOf
[...]

PR #264, opened 2022-08-18
Replaces instances of `anyOf` to `oneOf`

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/245

This PR: 
- Replaces `anyOf` to `oneOf` for geometric points.
- Replaces instances of `anyOf` to `oneOf` for `Decimal`, `Double`, `Single` and `Int64` `EdmPrimitiveTypeKind`.
- Updates unit and integration tests.
- Make components for floating point numbers.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
@@ -183,11 +183,13 @@ namespace Microsoft.OpenApi.OData.Generator
                         schema.Type = "number";
                         schema.Format = format;
                     }
+                    schema.Format = format;
                     break;
                 case EdmPrimitiveTypeKind.Double: // double
+                    format = "double";
                     schema.OneOf = new List<OpenApiSchema>
                     {
-                        new OpenApiSchema { Type = "number", Format = "double" },
+                        new OpenApiSchema { Type = "number", Format = format },
                         new OpenApiSchema { Type = "string" },
                         new OpenApiSchema
                         {
@@ -199,11 +201,13 @@ namespace Microsoft.OpenApi.OData.Generator
                             }
                         }
                     };
+                    schema.Format = format;
                     break;
                 case EdmPrimitiveTypeKind.Single: // single
+                    format = "float";
                     schema.OneOf = new List<OpenApiSchema>
                     {
-                        new OpenApiSchema { Type = "number", Format = "float"},
+                        new OpenApiSchema { Type = "number", Format = format },
                         new OpenApiSchema { Type = "string" },
                         new OpenApiSchema
                         {
@@ -215,6 +219,7 @@ namespace Microsoft.OpenApi.OData.Generator
                             }
                         }
                     };
+                    schema.Format = format;
                     break;
                 case EdmPrimitiveTypeKind.Guid: // guid
                     schema.Type = "string";
@@ -248,6 +253,7 @@ namespace Microsoft.OpenApi.OData.Generator
                         schema.Type = "integer";
                         schema.Format = format;
                     }
+                    schema.Format = format;
                     break;
                 case EdmPrimitiveTypeKind.SByte:
                     schema.Type = "integer";
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
@@ -169,7 +169,7 @@ namespace Microsoft.OpenApi.OData.Generator
                 case EdmPrimitiveTypeKind.Decimal: // decimal
                     if (context.Settings.IEEE754Compatible)
                     {
-                        schema.AnyOf = new List<OpenApiSchema>
+                        schema.OneOf = new List<OpenApiSchema>
                         {
                             new OpenApiSchema { Type = "number" },
                             new OpenApiSchema { Type = "string" },
@@ -182,7 +182,7 @@ namespace Microsoft.OpenApi.OData.Generator
                     schema.Format = "decimal";
                     break;
                 case EdmPrimitiveTypeKind.Double: // double
-                    schema.AnyOf = new List<OpenApiSchema>
+                    schema.OneOf = new List<OpenApiSchema>
                     {
                         new OpenApiSchema { Type = "number" },
                         new OpenApiSchema { Type = "string" },
@@ -199,7 +199,7 @@ namespace Microsoft.OpenApi.OData.Generator
                     schema.Format = "double";
                     break;
                 case EdmPrimitiveTypeKind.Single: // single
-                    schema.AnyOf = new List<OpenApiSchema>
+                    schema.OneOf = new List<OpenApiSchema>
                     {
                         new OpenApiSchema { Type = "number" },
                         new OpenApiSchema { Type = "string" },
@@ -235,7 +235,7 @@ namespace Microsoft.OpenApi.OData.Generator
                 case EdmPrimitiveTypeKind.Int64:
                     if (context.Settings.IEEE754Compatible)
                     {
-                        schema.AnyOf = new List<OpenApiSchema>
+                        schema.OneOf = new List<OpenApiSchema>
                         {
                             new OpenApiSchema { Type = "integer" },
                             new OpenApiSchema { Type = "string" }
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
@@ -147,6 +147,8 @@ namespace Microsoft.OpenApi.OData.Generator
                 OneOf = null,
                 AnyOf = null
             };
+
+            string format;
             switch (primitiveType.PrimitiveKind)
             {
                 case EdmPrimitiveTypeKind.Binary: // binary
@@ -167,24 +169,25 @@ namespace Microsoft.OpenApi.OData.Generator
                     schema.Pattern = "^[0-9]{4,}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]([.][0-9]{1,12})?(Z|[+-][0-9][0-9]:[0-9][0-9])$";
                     break;
                 case EdmPrimitiveTypeKind.Decimal: // decimal
+                    format = "decimal";
                     if (context.Settings.IEEE754Compatible)
                     {
                         schema.OneOf = new List<OpenApiSchema>
                         {
-                            new OpenApiSchema { Type = "number" },
+                            new OpenApiSchema { Type = "number", Format = format },
                             new OpenApiSchema { Type = "string" },
                         };
                     }
                     else
                     {
                         schema.Type = "number";
+                        schema.Format = format;
                     }
-                    schema.Format = "decimal";
                     break;
                 case EdmPrimitiveTypeKind.Double: // double
                     schema.OneOf = new List<OpenApiSchema>
                     {
-                        new OpenApiSchema { Type = "number" },
+                        new OpenApiSchema { Type = "number", Format = "double" },
                         new OpenApiSchema { Type = "string" },
                         new OpenApiSchema
                         {
@@ -196,12 +199,11 @@ namespace Microsoft.OpenApi.OData.Generator
                             }
                         }
                     };
-                    schema.Format = "double";
                     break;
                 case EdmPrimitiveTypeKind.Single: // single
                     schema.OneOf = new List<OpenApiSchema>
                     {
-                        new OpenApiSchema { Type = "number" },
+                        new OpenApiSchema { Type = "number", Format = "float"},
                         new OpenApiSchema { Type = "string" },
                         new OpenApiSchema
                         {
@@ -213,7 +215,6 @@ namespace Microsoft.OpenApi.OData.Generator
                             }
                         }
                     };
-                    schema.Format = "float";
                     break;
                 case EdmPrimitiveTypeKind.Guid: // guid
                     schema.Type = "string";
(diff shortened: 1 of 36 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
```diff
@@ -182,34 +182,34 @@ public static OpenApiSchema CreateSchema(this ODataContext context, IEdmPrimitiv
                     schema.Format = "decimal";
```
baywet, line 182: I believe the format should be on the number schema

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
```diff
@@ -182,34 +182,34 @@ public static OpenApiSchema CreateSchema(this ODataContext context, IEdmPrimitiv
...
                             }
                         }
                     };
                     schema.Format = "double";
                     break;
                 case EdmPrimitiveTypeKind.Single: // single
-                    schema.AnyOf = new List<OpenApiSchema>
+                    schema.OneOf = new List<OpenApiSchema>
                     {
                         new OpenApiSchema { Type = "number" },
```
baywet, line 204: I believe we should also have format float

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
```diff
@@ -235,7 +235,7 @@ public static OpenApiSchema CreateSchema(this ODataContext context, IEdmPrimitiv
                 case EdmPrimitiveTypeKind.Int64:
                     if (context.Settings.IEEE754Compatible)
                     {
-                        schema.AnyOf = new List<OpenApiSchema>
+                        schema.OneOf = new List<OpenApiSchema>
                         {
                             new OpenApiSchema { Type = "integer" },
```
baywet, line 240: format int64
https://spec.openapis.org/oas/latest.html#data-types

baywet (changes requested):
Thanks for addressing this, a couple of remarks

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
baywet: in addition to setting the type, the format should be set to int64

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
baywet: instead of having the format on the main schema object, it should be set to the schema that sets the type to number

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -576,7 +586,29 @@ private static IOpenApiAny GetTypeNameForExample(ODataContext context, IEdmTypeR
...
+                                {
+                                    schemaList = schemaList.Union(schema.OneOf).ToList();
+                                }
+
+                                if (schema.AllOf != null)
+                                {
+                                    schemaList = schemaList.Union(schema.AllOf).ToList();
+                                }
+
+                                typeName = schemaList?.FirstOrDefault(x => !string.IsNullOrEmpty(x.Format))?.Format;
```
baywet, line 608:
```suggestion
                                (schema.AnyOf ?? Enumerable.Empty<OpenApiSchema>())
                                .Union(schema.AllOf ?? Enumerable.Empty<OpenApiSchema>())
                                .Union(schema.OneOf ?? Enumerable.Empty<OpenApiSchema>())
                                .FirstOrDefault(static x => !string.IsNullOrEmpty(x.Format))?.Format;                                
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
```diff
@@ -147,6 +147,8 @@ public static OpenApiSchema CreateSchema(this ODataContext context, IEdmPrimitiv
                 OneOf = null,
                 AnyOf = null
             };
+
+            string format;
```
baywet, line 151: format is never being assigned?

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
irvinesunday: They are being used within the use case blocks: https://github.com/microsoft/OpenAPI.NET.OData/blob/5e64822e4d90dda71e8d86571e3cb7af3cb99e5b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs#L172

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -576,7 +586,16 @@ private static IOpenApiAny GetTypeNameForExample(ODataContext context, IEdmTypeR
                         }
                         else
                         {
-                            return new OpenApiString(schema.Type ?? schema.Format);
+                            string typeName = schema.Type ?? schema.Format;
+                            if (typeName == null)
+                            {
+                                (schema.AnyOf ?? Enumerable.Empty<OpenApiSchema>())
```
baywet, line 592:
```suggestion
                                typeName = (schema.AnyOf ?? Enumerable.Empty<OpenApiSchema>())
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -576,7 +586,16 @@ private static IOpenApiAny GetTypeNameForExample(ODataContext context, IEdmTypeR
                         }
                         else
                         {
-                            return new OpenApiString(schema.Type ?? schema.Format);
+                            string typeName = schema.Type ?? schema.Format;
+                            if (typeName == null)
```
baywet, line 590:
```suggestion
                            if (string.IsNullOrEmpty(typeName))
```

baywet (changes requested):
sorry I missed the affectation

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
```diff
@@ -25794,9 +25792,7 @@
         "Gender": {
           "$ref": "#/definitions/Microsoft.OData.Service.Sample.TrippinInMemory.Models.PersonGender"
         },
-        "Age": {
-          "format": "int64"
-        },
+        "Age": { },
```
baywet, line 25795: now I understand why the format was set on the root object. for v2....
the serialize as v2 probably needs some further fixing to make sure the int64 and double formats are present.

baywet (changes requested):

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
irvinesunday: Could this be a potential challenge since v2 doesn't support `anyOf` and `oneOf `constructs?

baywet: To be specific: I'm not saying we should have anyof/oneof in v2 as you mentioned it's not supported. What I'm requesting is to have the type/format set on the root schema (extracted from the any/oneof at serialization time). Because with the current code, we're losing information over the previous version.

irvinesunday: Do we still want to retain the `Format` properties defined within the respective schemas of the `oneOf`?

baywet: in case of v2 only we want to retain the format and type from one of the oneof/anyof. In case of v3, since we can have a more precise description, the root node shouldn't have any format/type, but the sub-schemas will contain the information. I hope that clarifies the point.

baywet: Right, but we still don't want the format to appear on the root schema object when serializing to v3, only for v2

irvinesunday: Seems like this should be resolved in the OpenAPI.NET lib. I don't see a way to include the `Format` property in the root of the schema object for v2 and exclude it for v3.

baywet: right, would you mind opening a PR on the openapi.net lib to tackle the problem end to end please?

irvinesunday: Here's the issue linked to that effect: https://github.com/microsoft/OpenAPI.NET/issues/992
Shall we proceed to merge this PR in the meantime, or is this fix required before proceeding?

baywet: Thanks for creating the issue. I believe we should fix the OpenAPI.net library first since otherwise we'll ship a regression. I'll add my thoughts on the issue over there.

baywet: preview is released, can you give it a try? https://www.nuget.org/packages/Microsoft.OpenApi/1.4.0-preview3

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
```diff
@@ -1167,14 +1167,12 @@
           {
             "in": "path",
             "name": "lat",
-            "required": true,
-            "format": "double"
+            "required": true
```
baywet, line 1170: the format of this one is still missing. Maybe something was missed in the OpenAPI.Net lib?
Other than that, the rest of v2 and v3 looks good.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
```diff
@@ -167,53 +169,54 @@ public static OpenApiSchema CreateSchema(this ODataContext context, IEdmPrimitiv
                     schema.Pattern = "^[0-9]{4,}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9]([.][0-9]{1,12})?(Z|[+-][0-9][0-9]:[0-9][0-9])$";
                     break;
                 case EdmPrimitiveTypeKind.Decimal: // decimal
+                    format = "decimal";
```
baywet, line 172: as far as I can tell the format variable is only used in the switch case where it's assigned, maybe we could get rid of the variable and use the object initialization directly for sanity now?

baywet (changes requested):

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
irvinesunday: That's because we in Parameter serialization for v2, we aren't serializing the schema inside this condition: https://github.com/microsoft/OpenAPI.NET/blob/eb4778d1caf753eab32ce71517b3a02b89a7d46f/src/Microsoft.OpenApi/Models/OpenApiParameter.cs#L333-L335

This means that we are also not getting the oneOf/allOff composition into oneOf.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
irvinesunday: The variable used here is to avoid duplication of the format values within those switch statement blocks

baywet: it then should be a local constant instead of a variable to avoid affectations

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
baywet: And I believe this should be addressed since the format property is supported on parameters.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiEdmTypeSchemaGenerator.cs
irvinesunday: Okay, I agree.

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
irvinesunday: You were right, we had missed something in the fix that was done in the OpenAPI.NET. Here's the PR to remedy that: https://github.com/microsoft/OpenAPI.NET/pull/1001

baywet: thanks for making the fix. resolving

baywet:
I think some unit tests still need to be updated and we should be good to go after that.

baywet (approved):

Issue #196, opened 2022-03-11
Add support for @odata.context and @odata.count


sherlock1982:
We have paging support and can specify `$count=true` to get the count of all elements. 
Unfortunately there's no way to get `@odata.count` value from a response.

Without this OData paging features are kind of unusable.
In order to add it I modified the document approximately like this:

            foreach (var path in document.Paths.Values)
            {
                foreach (var operation in path.Operations.Values)
                {
                    if (operation.OperationId.Contains(".List", StringComparison.Ordinal))
                    {
                        var props = operation.Responses["200"].Content["application/json"].Schema.Properties;
                        props.Add("@odata.context", new OpenApiSchema()
                        {
                            Type = "string"
                        });
                        props.Add("@odata.count", new OpenApiSchema()
                        {
                            Type = "integer",
                            Format = "int32"
                        });
                    }
               }
           }

darrelmiller:
@sherlock1982 Do you have a need for @odata.context? I understand how count and nextlink are useful but to date the only tooling that I have seen that uses @odata.context is native OData tooling and therefore does not seem useful in an OpenAPI description.

sherlock1982:
It's ok without `@odata.context`. 
Thought to add it just because it exists :-)

irvinesunday:
We already have a setting `context.Settings.EnablePagination` that enables `@odata.nextLink` in collection responses. We can append `@odata.count` to this as well.

PR #270, opened 2022-08-31
Adds `@odata.count` parameter to collection responses

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/196

This PR:
- Adds an `@odata.count` as a pagination property for collection responses. Together with `@odata.nextLink` these are componentized and reused.
- Adds a new setting `Settings.EnableCount` which controls whether to set the `@odata.count`.
- When `Settings.EnablePagination` and `Settings.EnableCount` are enabled, these properties alongside the base properties of a type are appended to an `allOf` schema.
- Updates test.

When `Settings.EnablePagination = true` and `Settings.EnableCount = true` the new schema for collections would look like:

image

And the `BaseCollectionPaginationCountResponse` schema:

image

And the `BaseCollectionCountResponse` schema:

image

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
@@ -106,16 +106,6 @@ namespace Microsoft.OpenApi.OData.Common
         public static string CollectionSchemaSuffix = "CollectionResponse";
 
         /// <summary>
-        /// Suffix used for the base collection pagination response schema.
-        /// </summary>
-        public static string BaseCollectionPaginationResponse = "BaseCollectionPaginationResponse";
-
-        /// <summary>
-        /// Suffix used for the base collection count response schema.
-        /// </summary>
-        public static string BaseCollectionCountResponse = "BaseCollectionCountResponse";
-
-        /// <summary>
         /// Suffix used for the base collection pagination response schema and count response schemas.
         /// </summary>
         public static string BaseCollectionPaginationCountResponse = "BaseCollectionPaginationCountResponse";
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
@@ -106,10 +106,14 @@ namespace Microsoft.OpenApi.OData.Common
         public static string CollectionSchemaSuffix = "CollectionResponse";
 
         /// <summary>
+        /// Suffix used for collection pagination response schemas.
+        /// </summary>
+        public static string CollectionPaginationResponse = "CollectionPaginationResponse";
+
+        /// <summary>
         /// Name used for reference update.
         /// </summary>
         public static string ReferenceUpdateSchemaName = "ReferenceUpdate";
-
 
         /// <summary>
         /// Name used for reference update.
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
@@ -116,6 +116,11 @@ namespace Microsoft.OpenApi.OData.Common
         public static string BaseCollectionCountResponse = "BaseCollectionCountResponse";
 
         /// <summary>
+        /// Suffix used for the base collection pagination response schema and count response schemas.
+        /// </summary>
+        public static string BaseCollectionPaginationCountResponse = "BaseCollectionPaginationCountResponse";
+
+        /// <summary>
         /// Name used for reference update.
         /// </summary>
         public static string ReferenceUpdateSchemaName = "ReferenceUpdate";
@@ -144,5 +149,15 @@ namespace Microsoft.OpenApi.OData.Common
         /// The odata id.
         /// </summary>
         public static string OdataId = "@odata.id";
+
+        /// <summary>
+        /// object type
+        /// </summary>
+        public static string ObjectType = "object";
+
+        /// <summary>
+        /// string type
+        /// </summary>
+        public static string StringType = "string";
     }
 }
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
@@ -106,9 +106,14 @@ namespace Microsoft.OpenApi.OData.Common
         public static string CollectionSchemaSuffix = "CollectionResponse";
 
         /// <summary>
-        /// Suffix used for collection pagination response schemas.
+        /// Suffix used for the base collection pagination response schema.
         /// </summary>
-        public static string CollectionPaginationResponse = "CollectionPaginationResponse";
+        public static string BaseCollectionPaginationResponse = "BaseCollectionPaginationResponse";
+
+        /// <summary>
+        /// Suffix used for the base collection count response schema.
+        /// </summary>
+        public static string BaseCollectionCountResponse = "BaseCollectionCountResponse";
 
         /// <summary>
         /// Name used for reference update.
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
@@ -95,53 +95,61 @@ namespace Microsoft.OpenApi.OData.Generator
             
             if(context.HasAnyNonContainedCollections())                                        
             {
-                schemas[$"String{Constants.CollectionSchemaSuffix}"] = CreateCollectionSchema(context, new OpenApiSchema { Type = "string" }, "string");
+                schemas[$"String{Constants.CollectionSchemaSuffix}"] = CreateCollectionSchema(context, new OpenApiSchema { Type = Constants.StringType }, Constants.StringType);
             }
 
             schemas[Constants.ReferenceUpdateSchemaName] = new()
             {
-                Type = "object",
+                Type = Constants.ObjectType,
                 Properties = new Dictionary<string, OpenApiSchema>
                     {
-                        {Constants.OdataId, new OpenApiSchema { Type = "string", Nullable = false }},
-                        {Constants.OdataType, new OpenApiSchema { Type = "string", Nullable = true }},
+                        {Constants.OdataId, new OpenApiSchema { Type = Constants.StringType, Nullable = false }},
+                        {Constants.OdataType, new OpenApiSchema { Type = Constants.StringType, Nullable = true }},
                     }
             };
 
             schemas[Constants.ReferenceCreateSchemaName] = new()
             {
-                Type = "object",
+                Type = Constants.ObjectType,
                 Properties = new Dictionary<string, OpenApiSchema>
                 {
-                    {Constants.OdataId, new OpenApiSchema { Type = "string", Nullable = false }}
+                    {Constants.OdataId, new OpenApiSchema { Type = Constants.StringType, Nullable = false }}
                 },
-                AdditionalProperties = new OpenApiSchema { Type = "object" }
+                AdditionalProperties = new OpenApiSchema { Type = Constants.ObjectType }
             };
 
+            // @odata.nextLink
             if (context.Settings.EnablePagination)
             {
                 schemas[Constants.BaseCollectionPaginationResponse] = new()
                 {
                     Title = "Base collection pagination response",
-                    Type = "object",
-                    Properties = new Dictionary<string, OpenApiSchema>
-                    {
-                        { "@odata.nextLink", new OpenApiSchema { Type = "string"} }
-                    }
+                    Type = Constants.ObjectType
                 };
+                schemas[Constants.BaseCollectionPaginationResponse].Properties.Add(ODataConstants.OdataNextLink);
             }
 
+            // @odata.count
             if (context.Settings.EnableCount)
             {
                 schemas[Constants.BaseCollectionCountResponse] = new()
                 {
                     Title = "Base collection count response",
-                    Type = "object",
-                    Properties = new Dictionary<string, OpenApiSchema>
-                    {
-                        { "@odata.count", new OpenApiSchema { Type = "integer", Format = "int64" } }
-                    }
+                    Type = Constants.ObjectType
                 };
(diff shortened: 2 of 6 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -118,6 +118,20 @@ internal static class OpenApiSchemaGenerator
...
 
+            if (context.Settings.EnablePagination)
+            {
+                schemas[Constants.CollectionPaginationResponse] = new()
+                {
+                    Title = "Base collection response",
+                    Type = "object",
+                    Properties = new Dictionary<string, OpenApiSchema>
+                    {
+                        { "@odata.nextLink", new OpenApiSchema { Type = "string"} },
```
baywet, line 129: should we set them both as nullable? They are not always present

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -198,22 +212,41 @@ private static OpenApiSchema CreateCollectionSchema(ODataContext context, OpenAp
...
+                            {
+                                Type = ReferenceType.Schema,
+                                Id = Constants.CollectionPaginationResponse
+                            }
+                        },
+                        baseSchema
+                    }
+                };
+            }
+            else
```
baywet, line 242: I wonder whether the Count property should always be here? Regardless of the pagination setting.
To achieve that we could split the base schema in two BaseCollectionCountResponse and BaseCollectionPagination response. The later is an all of with the former.
Then here depending on the setting, it's either an allOf count response + inline (pagination disabled) or pagination response + inline (pagination enabled).
What do you think?

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: I agree. They are inherently independent, and need to be able to be set and referenced independently.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
baywet: sure, but pagination can be enabled from a translation perspective and the service still not reply with a next link and/or count at runtime.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -198,22 +224,76 @@ private static OpenApiSchema CreateCollectionSchema(ODataContext context, OpenAp
...
+            };
+
+            OpenApiSchema colSchema;
+            if (context.Settings.EnablePagination && context.Settings.EnableCount)
+            {
+                colSchema = new OpenApiSchema
+                {
+                    AllOf = new List<OpenApiSchema>
+                    {
+                        paginationSchema,
```
baywet, line 261: this one with 3 allOf is going to be challenging since it suggests multi-inheritance, which most languages don't support.
My suggestion it to make it easier for code generators by making count an all of inline count definition + reference to pagination (or vice versa) when both settings are enabled.
And here we'd only be left with the last two blocks and the following condition for the first block
`context.Settings.EnablePagination && !context.Settings.EnableCount`
What do you think?

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: Or alternatively, we can have a 3rd `BaseCollectionPaginationCountResponse` schema that combines the two. When both settings are `true` then we use the above reference schema, else we use the individual respective schemas.

baywet: and that new base would redefine at least one property, not be a schema with 2 referenced all of entries right?

irvinesunday: The entity collection would look like this:

```
microsoft.graph.userCollectionResponse:
  title: Collection of user
  type: object
  allOf:
    - $ref: '#/components/schemas/BaseCollectionPaginationCountResponse'
    - type: object
      properties:
        value:
          type: array
          items:
            $ref: '#/components/schemas/microsoft.graph.user'
```

and the new combined base `BaseCollectionPaginationCountResponse` would look like:

```
BaseCollectionPaginationCountResponse:
  title: Base collection pagination and count responses
  type: object
  properties:
    '@odata.nextLink':
      type: string
    '@odata.count':
      type: integer
      format: int64
```
So in total, we have 3 collection bases:
1. Pagination
2. Count
3. Pagination + Count

Each will be defined depending on the settings

baywet: alright, let's roll with that. Thanks for the discussion.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: Resolved this with `nullable = true`

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -198,22 +232,88 @@ private static OpenApiSchema CreateCollectionSchema(ODataContext context, OpenAp
...
+                UnresolvedReference = true,
+                Reference = new OpenApiReference
+                {
+                    Type = ReferenceType.Schema,
+                    Id = Constants.BaseCollectionPaginationCountResponse
+                }
+            };
+
+            OpenApiSchema colSchema;
+            if (context.Settings.EnablePagination && context.Settings.EnableCount)
```
baywet, line 276:
```suggestion
            if (context.Settings.EnablePagination || context.Settings.EnableCount)
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -198,22 +232,88 @@ private static OpenApiSchema CreateCollectionSchema(ODataContext context, OpenAp
...
+            {
+                colSchema = new OpenApiSchema
+                {
+                    AllOf = new List<OpenApiSchema>
+                    {
+                        countSchema,
+                        baseSchema
+                    }
+                };
+            }
```
baywet, line 308:
```suggestion
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -198,22 +232,88 @@ private static OpenApiSchema CreateCollectionSchema(ODataContext context, OpenAp
...
+            // @odata.count
+            OpenApiSchema countSchema = new()
+            {
+                UnresolvedReference = true,
+                Reference = new OpenApiReference
+                {
+                    Type = ReferenceType.Schema,
+                    Id = Constants.BaseCollectionCountResponse
+                }
+            };
```
baywet, line 262:
```suggestion
```

src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
```diff
@@ -105,12 +105,26 @@ internal static class Constants
...
+        /// <summary>
+        /// Suffix used for the base collection pagination response schema.
+        /// </summary>
+        public static string BaseCollectionPaginationResponse = "BaseCollectionPaginationResponse";
+
+        /// <summary>
+        /// Suffix used for the base collection count response schema.
+        /// </summary>
+        public static string BaseCollectionCountResponse = "BaseCollectionCountResponse";
+
```
baywet, line 117:
```suggestion
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -95,29 +95,63 @@ internal static class OpenApiSchemaGenerator
...
+            if (context.Settings.EnableCount)
+            {
+                schemas[Constants.BaseCollectionCountResponse] = new()
+                {
+                    Title = "Base collection count response",
+                    Type = Constants.ObjectType
+                };
+                schemas[Constants.BaseCollectionCountResponse].Properties.Add(ODataConstants.OdataCount);
+            }
+
```
baywet, line 142:
```suggestion
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -95,29 +95,63 @@ internal static class OpenApiSchemaGenerator
...
+            // @odata.nextLink + @odata.count
+            if (context.Settings.EnablePagination && context.Settings.EnableCount)
```
baywet, line 144:
```suggestion
            if (context.Settings.EnablePagination || context.Settings.EnableCount)
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -95,29 +95,63 @@ internal static class OpenApiSchemaGenerator
...
+            {
+                schemas[Constants.BaseCollectionPaginationCountResponse] = new()
+                {
+                    Title = "Base collection pagination and count responses",
+                    Type = Constants.ObjectType,                   
+                };
+                schemas[Constants.BaseCollectionPaginationCountResponse].Properties.Add(ODataConstants.OdataCount);
+                schemas[Constants.BaseCollectionPaginationCountResponse].Properties.Add(ODataConstants.OdataNextLink);
```
baywet, line 152:
```suggestion
                if (context.Settings.EnableCount) schemas[Constants.BaseCollectionPaginationCountResponse].Properties.Add(ODataConstants.OdataCount);
     if (context.Settings.EnablePagination) schemas[Constants.BaseCollectionPaginationCountResponse].Properties.Add(ODataConstants.OdataNextLink);
```

baywet (changes requested):

baywet (approved):
Thanks for making the changes, I pushed a much simpler implementation :)

test/Microsoft.OpenAPI.OData.Reader.Tests/Generator/OpenApiSchemaGeneratorTests.cs
```diff
@@ -35,28 +35,51 @@ public void CreateSchemasThrowArgumentNullContext()
             Assert.Throws<ArgumentNullException>("context", () => context.CreateSchemas());
         }
 
-        [Fact]
-        public void CreatesCollectionResponseSchema()
+        [Theory]
+        [InlineData(true, false, "BaseCollectionPaginationResponse")]
+        [InlineData(false, true, "BaseCollectionCountResponse")]
+        [InlineData(true, true, "BaseCollectionPaginationCountResponse")]
```
baywet, line 41:
```suggestion
        [InlineData(true, false, "BaseCollectionPaginationCountResponse")]
        [InlineData(false, true, "BaseCollectionPaginationCountResponse")]
        [InlineData(true, true, "BaseCollectionPaginationCountResponse")]
```

baywet (changes requested):

baywet (approved):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: What about when both settings are set to `true`?

baywet: the base type will have both properties in that case

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: Same question as above, what about when both settings are set to `true`?

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: If in either case, both properties are added, then why have the 2 different settings? Why not just have 1?

baywet: because as you mentioned earlier, you can want count to be part of your conversion, paging, or both. They are not overlapping or mutually exclusive.

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
baywet: Same answer as above :)

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: This was borne from your earlier comment :)
https://github.com/microsoft/OpenAPI.NET.OData/pull/270#discussion_r959517460

baywet: yep, thanks for double checking. I do contradict myself sometimes. But I think we have a flexible and consistent solution which is low maintenance right now. Unless you disagree, we're good to merge IMHO.

irvinesunday: I'm good with it.

Issue #259, opened 2022-08-02
Semantic violation: Certain OData functions have path parameters referenced in query


peombwa:
Certain OData functions have path parameters referenced in `query` instead of `path`. This results in generation error since code generators cannot find the specified path parameters.

#### Current Representation
``` yml
'/users/{user-id}/joinedTeams/{team-id}/primaryChannel/microsoft.graph.doesUserHaveAccess(userId=''{userId}'',tenantId=''{tenantId}'',userPrincipalName=''{userPrincipalName}'')':
  get:
    tags:
      - users.Functions
    summary: Invoke function doesUserHaveAccess
    operationId: users.joinedTeams.primaryChannel_doesUserHaveAccess
    parameters:
      - name: user-id
        in: path
        description: 'key: id of user'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: user
      - name: team-id
        in: path
        description: 'key: id of team'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: team
      - name: userId
        in: query
        description: 'Usage: userId=''{userId}'''
        schema:
          type: string
          nullable: true
      - name: tenantId
        in: query
        description: 'Usage: tenantId=''{tenantId}'''
        schema:
          type: string
          nullable: true
      - name: userPrincipalName
        in: query
        description: 'Usage: userPrincipalName=''{userPrincipalName}'''
        schema:
          type: string
          nullable: true
    responses:
      '200': ...
      default:
        $ref: '#/components/responses/error'
    x-ms-docs-operation-type: function
```

#### Ideal Representation
``` diff
'/users/{user-id}/joinedTeams/{team-id}/primaryChannel/microsoft.graph.doesUserHaveAccess(userId=''{userId}'',tenantId=''{tenantId}'',userPrincipalName=''{userPrincipalName}'')':
  get:
    tags:
      - users.Functions
    summary: Invoke function doesUserHaveAccess
    operationId: users.joinedTeams.primaryChannel_doesUserHaveAccess
    parameters:
      - name: user-id
        in: path
        description: 'key: id of user'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: user
      - name: team-id
        in: path
        description: 'key: id of team'
        required: true
        schema:
          type: string
        x-ms-docs-key-type: team
      - name: userId
+         in: path
-         in: query
        description: 'Usage: userId=''{userId}'''
        schema:
          type: string
          nullable: true
      - name: tenantId
[...]

PR #272, opened 2022-09-05
Alias optional parameters of OData functions

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/259

This PR:
- Aliases optional parameters of OData functions by appending `@` together with the parameter name to denote the parameter alias.
- Updates tests to validate the above

Paths of functions with optional parameters will look like below:
 
image

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data.Common;
 using System.Linq;
 using System.Text;
 using Microsoft.OData.Edm;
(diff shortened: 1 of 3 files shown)


baywet (changes requested):
part of the issue what that path parameters were described as in: query. And I haven't seen any code change related to that aspect. A miss?

irvinesunday:
> part of the issue what that path parameters were described as in: query. And I haven't seen any code change related to that aspect. A miss?

We don't describe path parameters as in: `query`

Given this function example:

image

 And this path which is bound to that function
 
image
You can see that the path parameters: `team-id` and `channel-id` are defined as `in: path` whereas the function parameters: `userId`, `tenantId` and `userPrincipalName` are all defined as `in: query`

baywet:
what happens for required parameters? I believe the fact they were present in the path but documented to be part of query was the original defect reported by @peombwa

irvinesunday:
> what happens for required parameters? I believe the fact they were present in the path but documented to be part of query was the original defect reported by @peombwa

Given this function which has a required parameter `userid`

image

And the subsequently generated path...

image

You can see that the required parameter `userid` is generated as `in: path` and with the correct syntax in the path.

The function that @peombwa was referencing in the target issue had all optional parameters:

image

baywet (approved):
Thanks for the clarification

Issue #274, opened 2022-09-06
explode for top/skip/filter/count/search query parameters should be set to false


baywet:
```diff
parameters:
    top:
      name: $top
      in: query
+    explode: false
      description: Show only the first n items
      schema:
        minimum: 0
        type: integer
      example: 50
    skip:
      name: $skip
      in: query
+    explode: false
      description: Skip the first n items
      schema:
        minimum: 0
        type: integer
    filter:
      name: $filter
      in: query
+    explode: false
      description: Filter items by property values
      schema:
        type: string
    count:
      name: $count
      in: query
+    explode: false
      description: Include count of items
      schema:
        type: boolean
    search:
      name: $search
      in: query
+    explode: false
      description: Search items by search phrases
      schema:
        type: string
```

I think it was missed when #107 was implemented. Select, orderby are compliant

PR #275, opened 2022-09-07
Set `explode = false` and `style = form` for OData query options

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/274

This PR:
- Sets `explode = false` and `style = form` for the OData query options: `$top`, `$skip`, `$search`, `$filter`, `$count`
- Update tests to validate the above.
- Update integration test files.


baywet (approved):
Thanks for tackling this so fast!

PR #291, opened 2022-10-03
Set assembly version used for conversion in OpenApiInfo object

### LINKED ISSUE #208: Provide a way to get the assembly version that was used when generating a file
It would be useful to provide an easy way to get the assembly version of the library that was used in generating an OpenAPI file e.g. in a comment in the generated file. This is useful for bug reporting and traceability.

millicentachieng:
Fixes #208

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiInfoGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiInfoGenerator.cs
@@ -3,9 +3,13 @@
 //  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // ------------------------------------------------------------
 
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
 using Microsoft.OData.Edm;
+using Microsoft.OpenApi.Any;
+using Microsoft.OpenApi.Interfaces;
 using Microsoft.OpenApi.Models;
 using Microsoft.OpenApi.OData.Common;
 using Microsoft.OpenApi.OData.Edm;
@@ -32,7 +36,8 @@ namespace Microsoft.OpenApi.OData.Generator
             {
                 Title = context.GetTitle(),
                 Version = context.GetVersion(),
-                Description = context.GetDescription()
+                Description = context.GetDescription(),
+                Extensions = context.GetExtensions()
             };
         }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiInfoGenerator.cs
```diff
@@ -93,5 +98,26 @@ private static string GetDescription(this ODataContext context)
...
+
+            if (context.Settings.ShowAssemblyInfo)
+            {
+                return new Dictionary<string, IOpenApiExtension>()
+                {
+                    {
+                        "x-generatedBy",
+                        new OpenApiObject
+                        {
+                            { "toolName", new OpenApiString("OpenAPI.NET.CSDL") },
```
irvinesunday, line 114: I think we should use the officially recognized nuget name for this library `Microsoft.OpenApi.OData`

src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReleaseNotes>
 - Use convert setting to toggle between referencing @odata.count and @odata.nextLink #282
 - Fixes URL Path parameters of type datetime generated as strings with quotes #262
+- Set assembly version used for conversion in OpenApiInfo object #208
```
irvinesunday, line 26: Update the version number as well: `1.2.0`

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiInfoGenerator.cs
```diff
@@ -93,5 +98,26 @@ private static string GetDescription(this ODataContext context)
...
+        private static Dictionary<string, IOpenApiExtension> GetExtensions(this ODataContext context)
+        {
+            Debug.Assert(context != null);
+
+            if (context.Settings.ShowAssemblyInfo)
+            {
+                return new Dictionary<string, IOpenApiExtension>()
+                {
+                    {
+                        "x-generatedBy",
```
baywet, line 111: Is this extension defined anywhere? Otherwise should we add a -ms and normalize to kebab casing?

src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
baywet: Preview 3 no?

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -275,6 +275,12 @@ public string PathPrefix
         /// </summary>
         public bool UseSuccessStatusCodeRange { get; set; } = false;
 
+        /// <summary>
+        /// Gets/Sets a value indicating whether to show the version of the assembly used for generating 
+        /// Open API document
+        /// </summary>
+        public bool ShowAssemblyInfo { get; set; } = true;
```
baywet, line 282: What do you think about include or add assembly information instead? Because it's not a UI setting, nothing else is going to "show"

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiInfoGenerator.cs
millicentachieng: I can't find it. I'll rename it.

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
millicentachieng: Yes, `IncludeAssemblyInfo` makes more sense.

baywet (dismissed):

baywet (approved):

Issue #265, opened 2022-08-23
Add support for `InsertRestrictions` and `UpdateRestrictions` on non-containment navigation properties


irvinesunday:
The generation of POST and PUT/PATCH operations on non-containment navigation properties should be done only when the above restrictions are explicitly set to `true`

PR #292, opened 2022-10-03
Adds create and update operations on non-containment nav. props when restrictions annotations are explicitly set to true

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/265

Proposes:
- Adding support for update and delete operations on non-containment navigation properties when restrictions annotations are explicitly set to true.
- Updates integration files.
- Updates release notes.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
+++ b/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
@@ -15,7 +15,7 @@
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <PackageId>Microsoft.OpenApi.OData</PackageId>
     <SignAssembly>true</SignAssembly>
-    <Version>1.2.0</Version>
+    <Version>1.2.0-preview3</Version>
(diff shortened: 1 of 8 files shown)


src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
```diff
@@ -15,14 +15,15 @@
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <PackageId>Microsoft.OpenApi.OData</PackageId>
     <SignAssembly>true</SignAssembly>
-    <Version>1.2.0-preview2</Version>
+    <Version>1.2.0</Version>
```
baywet, line 18: let's still use a preview version (4or 5?) so we can do some quality work with one weekly generation before we release the changes.

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
irvinesunday: Going with `-preview3`

baywet (approved):

Issue #269, opened 2022-08-30
`ExpandDerivedTypesNavigationProperties` setting granularity support


andrueastman:
Due to previous issues with the graph surface size seen in https://github.com/microsoft/OpenAPI.NET/issues/866 the `ExpandDerivedTypesNavigationProperties` setting is currently disabled in generating the openAPI from the CSDL metadata. 

This however may lead to some missing paths like actions/functions bound on the derived types or convenient nav Properties just one level deep. 

- https://docs.microsoft.com/en-us/graph/api/ediscovery-caseexportoperation-getdownloadurl?view=graph-rest-beta
- https://docs.microsoft.com/en-us/graph/api/intune-apps-mobileappcontentfile-commit?view=graph-rest-1.0

We could possibly consider adding a "depth control" setting to to limit how deep the paths would go to compromise on this rather than a yes/no situation.

irvinesunday:
The last time I tested this by generating derived types on navigation properties by going just one level deep I ended up with a significant number of paths generated. 
Maybe we can figure out the possibility of having a custom annotation that we can attach to the derived types that require expansion?

millicentachieng:
Should we also consider having a list of entity types to expand in the settings?

irvinesunday:
> Should we also consider having a list of entity types to expand in the settings?

That should probably be configured in the metadata as an annotation, as this would ideally be handled by the workload when they are updating their API.

irvinesunday:
For context, at the moment, we generate about 7K paths for v1.0. Expanding the navigation properties of derived types for the single type cast segment only generates about 126K paths.

@darrelmiller @baywet

baywet:
> Should we also consider having a list of entity types to expand in the settings?

I don't think this would be the best approach here, because I'd require the code that does the conversion (eg hidi, other tools) to know about the content of the CSDL in advance, which is not always the case (hidi, again).

I believe this problem of over expansion of methods has multiple facets:
- type cast segments should only have the methods of the target cast type under them, not the ones from the base type as those can be accessed by simply not casting.
- methods should only be added to singletons, entity sets, entity sets entries, single contained navigation properties, collection contained navigation properties and collection contained navigation properties entries.
- I believe @darrelmiller introduced an annotation to support unbound methods, but I don't remember the specifics.

Then there's also the over expansion of navigation properties themselves:
- Contains target is wrong at a bunch of places and we should either ask workloads to fix it or "fix it in post".
[...]

PR #302, opened 2022-10-26
Expands navigation properties of derived types only if declaring navigation property is a containment

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/269

This PR:
- Fixes the logic of expanding the navigation properties of derived types only if the declaring navigation property is a containment. Initially, expansion covered both containment and non-containment navigation properties.
- Adds a check to ensure we don't add a derived type that has already been added in the path when expanding the type's navigation properties.
- Update tests and integration files to validate the above.

To fully resolve the aforementioned issue, the below convert settings need to be set to `true`:
- `ExpandDerivedTypesNavigationProperties`
- `AppendBoundOperationsOnDerivedTypeCastSegments`

With the above fix and the above settings set to `true`, the result, for the examples provided in the issue, will be:
- https://docs.microsoft.com/en-us/graph/api/ediscovery-caseexportoperation-getdownloadurl?view=graph-rest-beta
image

- https://docs.microsoft.com/en-us/graph/api/intune-apps-mobileappcontentfile-commit?view=graph-rest-1.0
image


andrueastman (dismissed):
👍🏼

andrueastman:
Thanks @irvinesunday

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -559,21 +559,34 @@ private void CreateTypeCastPaths(ODataPath currentPath, OpenApiConvertSettings c
                                 .OfType<IEdmStructuredType>()
                                 .ToArray();
 
-            foreach(var targetType in targetTypes)
+            foreach (var targetType in targetTypes)
             {
-                var castPath = currentPath.Clone();
                 var targetTypeSegment = new ODataTypeCastSegment(targetType);
 
+                if (currentPath.Segments.Contains(targetTypeSegment))
```
baywet, line 566: this will always evaluate to false since the target type segment is instantiated at the previous lines (ref compare will fail) and since a comparer is not provided (value compare not done)

baywet (changes requested):

irvinesunday:
Adding more information for context:

The table below shows the number of **beta** paths generated when the two convert settings `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` are enabled/disabled. 

|  Convert Setting        | ExpandDerivedTypesNavigationProperties = true      | ExpandDerivedTypesNavigationProperties = false |
| -----------| ----------- | ----------- |
|**AppendBoundOperationsOnDerivedTypeCastSegments = true** | 21,484     | 16,969       |
|**AppendBoundOperationsOnDerivedTypeCastSegments = false** | 19,342   | 14,846        |

@peombwa / @timayabi2020 we might need to verify the generated OpenAPI document with PowerShell to verify whether there are any modules that might break AutoREST due to the increase in size.

peombwa:
> Adding more information for context:
> 
> The table below shows the number of **beta** paths generated when the two convert settings `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` are enabled/disabled.
> 
> Convert Setting	ExpandDerivedTypesNavigationProperties = true	ExpandDerivedTypesNavigationProperties = false
> **AppendBoundOperationsOnDerivedTypeCastSegments = true**	21,484	16,969
> **AppendBoundOperationsOnDerivedTypeCastSegments = false**	19,342	14,846
> @peombwa / @timayabi2020 we might need to verify the generated OpenAPI document with PowerShell to verify whether there are any modules that might break AutoREST due to the increase in size.

Thanks @irvinesunday for the breakdown!

Since OData type cast is currently [not enabled in DevX API](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1228), we may see an increase of `~4500` paths in beta. Can we have an experimental deployment of DevX API with `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` set to `true` to facilitate our validation?

Also, is `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` set to `false` by default? i.e., maintains the current nav property expansion logic?

irvinesunday:
> > Adding more information for context:
> > The table below shows the number of **beta** paths generated when the two convert settings `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` are enabled/disabled.
> > Convert Setting	ExpandDerivedTypesNavigationProperties = true	ExpandDerivedTypesNavigationProperties = false
> > **AppendBoundOperationsOnDerivedTypeCastSegments = true**	21,484	16,969
> > **AppendBoundOperationsOnDerivedTypeCastSegments = false**	19,342	14,846
> > @peombwa / @timayabi2020 we might need to verify the generated OpenAPI document with PowerShell to verify whether there are any modules that might break AutoREST due to the increase in size.
> 
> Thanks @irvinesunday for the breakdown!
> 
> Since OData type cast is currently [not enabled in DevX API](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1228), we may see an increase of `~4500` paths in beta. Can we have an experimental deployment of DevX API with `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` set to `true` to facilitate our validation?
> 
> Also, is `ExpandDerivedTypesNavigationProperties` and `AppendBoundOperationsOnDerivedTypeCastSegments` set to `false` by default? i.e., maintains the current nav property expansion logic?

Yes @peombwa, both settings are currently set to `false` in DevX API (`ExpandDerivedTypesNavigationProperties` explicitly; `AppendBoundOperationsOnDerivedTypeCastSegments` implicitly by default). We can definitely update [devxapitest](https://devxapitest.azurewebsites.net/) with the bumped up version of the conversion lib. and configure both settings to true. We can use this for our validation.

baywet (approved):
I'm a bit worried by the increase in the number of paths here, but I think we need to validate it before making a final decision. It'd be interesting to have a comparison of the newly added paths and see whether the service actually understands them. Especially for "deep" paths.

baywet:
(sorry about the auto-merge)

Issue #304, opened 2022-11-02
Default presence of `odata.type` property in generated OpenAPI


andrueastman:
Currently, all models are generated by with a default value for the `@odata.type` property which is further propagated into generated models in Kiota. This shouldn't be an issue per se but various workloads seem to not support the existence of this property as seen in the numerous linked issues to https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/534#issue-899279074.

The proposal therefore is to have the `@odata.type` not set by default in the result openApi generation and have it set in the following conditions where type disambiguation is necessary as the API will need to know the type to infer and take advantage of the properties.

1. The type derives from an abstract type
2. One of its base/parent types is referenced as the type for a property in another type (except if the base is entity).
3. One of its base types is referenced as the type in an odata action in another type (except if the base is entity).

Possibly, the default behaviour could have the `@odata.type` property set/unset and then we would add a setting to enable this behaiour as this may not really be a valid ask for another odata API needing the conversion to be performed.

Related issues
- https://github.com/microsoftgraph/msgraph-sdk-php/issues/1006
- https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/495

irvinesunday:
@peombwa / @timayabi2020 is this a behaviour that we need enabled/disabled for PowerShell as well?

peombwa:
@irvinesunday, it looks like we don't expose `@odata.type` property on models though DevX API - https://graphexplorerapi.azurewebsites.net/openapi?operationIds=drives.drive.root.createUploadSession&title=Files&openapiversion=3&style=Powershell&graphVersion=v1.0,

PR #307, opened 2022-11-11
Adds support for configuring the `default` value of derived types' `@odata.type` property

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/304

This PR:
- Adds support for setting the `default` value of the `@odata.type` property in the schema of derived types.
- Adds a new convert setting for setting this `default` value. The default value is `true`, meaning the default value will always be emitted. If `false` the value will be set conditionally based on whether the derived type's base type is abstract (and not entity) and is referenced in the properties of a structural property or an action.
- Updates existing tests and adds a new test to validate the above.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Numerics;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Edm.Vocabularies;
 using Microsoft.OpenApi.Any;
@@ -229,12 +230,14 @@ namespace Microsoft.OpenApi.OData.Common
         /// <param name="model">The Edm model.</param>
         /// <param name="baseType">The base type of the target <see cref="IEdmStructuredType"/>.</param>
         /// <param name="structuredTypes">Optional: The IEnumerable of <see cref="IEdmStructuredType"/> to check against.</param>
+        /// <param name="complexTypes">Optional: The IEnumerable of <see cref="IEdmComplexType"/> to check against.</param>
         /// <param name="actions">Optional: The IEnumerable of <see cref="IEdmAction"/> to check against.</param>
         /// <returns>True if reference is found, otherwise False.</returns>
         internal static bool IsBaseTypeReferencedAsTypeInModel(
             this IEdmModel model,
             IEdmStructuredType baseType,
             IEnumerable<IEdmStructuredType> structuredTypes = null,
+            IEnumerable<IEdmComplexType> complexTypes = null,
             IEnumerable<IEdmAction> actions = null)
         {
             string baseTypeName = baseType?.FullTypeName();
@@ -247,22 +250,36 @@ namespace Microsoft.OpenApi.OData.Common
                         .Where(static y => !y.Name.Equals(Constants.EntityName, StringComparison.OrdinalIgnoreCase))
                         .OfType<IEdmStructuredType>();
 
+                complexTypes ??= model.GetAllElements()
+                    .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.TypeDefinition)
+                    .OfType<IEdmComplexType>();          
+
                 actions ??= model.GetAllElements()
                         .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.Action)
                         .OfType<IEdmAction>();
 
                 // Is base type referenced as a type in any property within a structured type
                 bool isReferencedInStructuredType = structuredTypes
-                    .Any(x => x.DeclaredProperties.Where(x => x.Type.TypeKind() == EdmTypeKind.Entity)
-                    .Any(x => x.Type.FullName().Equals(baseTypeName, StringComparison.OrdinalIgnoreCase)));
+                    .Any(x => x.DeclaredProperties.Where(y => y.Type.TypeKind() == EdmTypeKind.Entity ||
+                                                            y.Type.TypeKind() == EdmTypeKind.Collection ||
+                                                            y.Type.TypeKind() == EdmTypeKind.Complex)
+                    .Any(z => z.Type.FullName().Equals(baseTypeName, StringComparison.OrdinalIgnoreCase)));
                 if (isReferencedInStructuredType) return true;
+
+                // Is base type referenced as a type within a complex type
+                bool isReferencedInComplexType = complexTypes
+                    .Any(x => x.DeclaredProperties.Where(y => y.Type.TypeKind() == EdmTypeKind.Entity ||
+                                                            y.Type.TypeKind() == EdmTypeKind.Collection ||
+                                                            y.Type.TypeKind() == EdmTypeKind.Complex)
+                    .Any(z => z.Type.FullName().Equals(baseTypeName, StringComparison.OrdinalIgnoreCase)));
+                if (isReferencedInComplexType) return true;
 
                 // Is base type referenced as a type in any parameter in an action
                 bool isReferencedInAction = actions.Any(x => x.Parameters.Any(x => x.Type.FullName().Equals(baseTypeName, StringComparison.OrdinalIgnoreCase)));
                 if (isReferencedInAction) return true;
 
                 // Recursively check the base type
-                return model.IsBaseTypeReferencedAsTypeInModel(baseType.BaseType, structuredTypes, actions);
+                return model.IsBaseTypeReferencedAsTypeInModel(baseType.BaseType, structuredTypes, complexTypes, actions);
             }
 
             return false;
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -237,15 +237,14 @@ namespace Microsoft.OpenApi.OData.Common
             IEnumerable<IEdmStructuredType> structuredTypes = null,
             IEnumerable<IEdmAction> actions = null)
         {
-            const string Entity = "entity";
             string baseTypeName = baseType?.FullTypeName();
-            bool isBaseTypeEntity = Entity.Equals(baseTypeName?.Split('.').Last(), StringComparison.OrdinalIgnoreCase);
+            bool isBaseTypeEntity = Constants.EntityName.Equals(baseTypeName?.Split('.').Last(), StringComparison.OrdinalIgnoreCase);
 
             if (!string.IsNullOrEmpty(baseTypeName) && !isBaseTypeEntity)
             {
                 structuredTypes ??= model.GetAllElements()
                         .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.TypeDefinition)
-                        .Where(static y => !y.Name.Equals(Entity, StringComparison.OrdinalIgnoreCase))
+                        .Where(static y => !y.Name.Equals(Constants.EntityName, StringComparison.OrdinalIgnoreCase))
                         .OfType<IEdmStructuredType>();
 
                 actions ??= model.GetAllElements()
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -230,14 +230,12 @@ namespace Microsoft.OpenApi.OData.Common
         /// <param name="model">The Edm model.</param>
         /// <param name="baseType">The base type of the target <see cref="IEdmStructuredType"/>.</param>
         /// <param name="structuredTypes">Optional: The IEnumerable of <see cref="IEdmStructuredType"/> to check against.</param>
-        /// <param name="complexTypes">Optional: The IEnumerable of <see cref="IEdmComplexType"/> to check against.</param>
         /// <param name="actions">Optional: The IEnumerable of <see cref="IEdmAction"/> to check against.</param>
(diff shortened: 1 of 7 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+        /// <param name="structuredTypes">Optional: The IEnumerable of <see cref="IEdmStructuredType"/> to check against.</param>
+        /// <param name="actions">Optional: The IEnumerable of <see cref="IEdmAction"/> to check against.</param>
+        /// <returns>True if reference is found, otherwise False.</returns>
+        internal static bool IsBaseTypeReferencedAsTypeInModel(
+            this IEdmModel model,
+            IEdmStructuredType baseType,
+            IEnumerable<IEdmStructuredType> structuredTypes = null,
+            IEnumerable<IEdmAction> actions = null)
+        {
+            const string Entity = "entity";
```
baywet, line 240: const should be declared outside of the method to declare it once

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+            string baseTypeName = baseType?.FullTypeName();
+            bool isBaseTypeEntity = Entity.Equals(baseTypeName?.Split('.').Last(), StringComparison.OrdinalIgnoreCase);
+
+            if (!string.IsNullOrEmpty(baseTypeName) && !isBaseTypeEntity)
+            {
+                structuredTypes ??= model.GetAllElements()
+                        .Where(x => x.SchemaElementKind == EdmSchemaElementKind.TypeDefinition)
```
baywet, line 247:
```suggestion
                        .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.TypeDefinition)
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                        .Where(y => !y.Name.Equals(Entity, StringComparison.OrdinalIgnoreCase))
```
baywet, line 248:
```suggestion
                        .Where(static y => !y.Name.Equals(Entity, StringComparison.OrdinalIgnoreCase))
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                        .OfType<IEdmStructuredType>();
+
+                actions ??= model.GetAllElements()
+                        .Where(x => x.SchemaElementKind == EdmSchemaElementKind.Action)
```
baywet, line 252:
```suggestion
                        .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.Action)
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                        .OfType<IEdmAction>();
+
+                // Is base type referenced as a type in any property within a structured type
+                bool isReferencedInStructuredType = structuredTypes
+                    .Any(x => x.DeclaredProperties.Where(x => x.Type.TypeKind() == EdmTypeKind.Entity)
```
baywet, line 257: don't you get an error saying variable x is already in use?

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                    .Any(x => x.Type.FullName().Equals(baseTypeName)));
```
baywet, line 258:
```suggestion
                    .Any(x => x.Type.FullName().Equals(baseTypeName, StringComparison.OrdinalIgnoreCase)));
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,51 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                if (isReferencedInStructuredType) return true;
+
+                // Is base type referenced as a type in any parameter in an action
+                bool isReferencedInAction = actions.Any(x => x.Parameters.Any(x => x.Type.FullName().Equals(baseTypeName)));
```
baywet, line 262:
```suggestion
                bool isReferencedInAction = actions.Any(x => x.Parameters.Any(x => x.Type.FullName().Equals(baseTypeName, StringComparison.OrdinalIgnoreCase)));
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
```diff
@@ -566,10 +566,21 @@ internal static OpenApiSchema CreateSchemaTypeSchema(this ODataContext context,
 
                 if (context.Settings.EnableDiscriminatorValue)
                 {
+                    OpenApiString defaultValue = null;
+                    bool isBaseTypeEntity = "entity".Equals(structuredType.BaseType?.FullTypeName().Split('.').Last(), StringComparison.OrdinalIgnoreCase);
+                    bool isBaseTypeAbstractNonEntity = (structuredType.BaseType?.IsAbstract ?? false) && !isBaseTypeEntity;
+
+                    if (context.Settings.EnableDefaultValueForOdataTypeProperty ||
```
baywet, line 573: shouldn't this be && instead of || ??

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: Yes, good catch! It should be:
```
if (context.Settings.EnableDefaultValueForOdataTypeProperty ||
    (!context.Settings.EnableDefaultValueForOdataTypeProperty &&
    (isBaseTypeAbstractNonEntity ||   
  context.Model.IsBaseTypeReferencedAsTypeInModel(structuredType.BaseType))))
```

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
irvinesunday: Yes, good catch! It should be:
```
if (context.Settings.EnableDefaultValueForOdataTypeProperty ||
    (!context.Settings.EnableDefaultValueForOdataTypeProperty &&
    (isBaseTypeAbstractNonEntity ||   
  context.Model.IsBaseTypeReferencedAsTypeInModel(structuredType.BaseType))))
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: Nope

andrueastman: I believe what @baywet is saying here is since the anonymous function is not static, it would be better to create a distinction between the `x` in the nested function and the outer one.

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +222,50 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                        .Where(static y => !y.Name.Equals(Constants.EntityName, StringComparison.OrdinalIgnoreCase))
+                        .OfType<IEdmStructuredType>();
+
+                actions ??= model.GetAllElements()
+                        .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.Action)
+                        .OfType<IEdmAction>();
+
+                // Is base type referenced as a type in any property within a structured type
+                bool isReferencedInStructuredType = structuredTypes
+                    .Any(x => x.DeclaredProperties.Where(x => x.Type.TypeKind() == EdmTypeKind.Entity)
```
andrueastman, line 256:
```suggestion
```
I'm not sure about this check being here because of two things,
- This leaves out checking Complex types which could possibly have an abstract base yet is still a stuctured type.
- The property could be a collection property and I believe `x.Type.TypeKind()` will return `EdmTypeKind.Collection` in such a scenario.

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -222,5 +223,66 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+            if (!string.IsNullOrEmpty(baseTypeName) && !isBaseTypeEntity)
+            {
+                structuredTypes ??= model.GetAllElements()
+                        .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.TypeDefinition)
+                        .Where(static y => !y.Name.Equals(Constants.EntityName, StringComparison.OrdinalIgnoreCase))
+                        .OfType<IEdmStructuredType>();
+
+                complexTypes ??= model.GetAllElements()
+                    .Where(static x => x.SchemaElementKind == EdmSchemaElementKind.TypeDefinition)
+                    .OfType<IEdmComplexType>();
```
andrueastman, line 255:
```suggestion       
```

One last thing, I think we don't need this now as `IEdmStructuredType` is implemented by `IEdmComplexType`. So, the `structuredTypes` collection should already include complex types. 


https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Schema/Interfaces/IEdmComplexType.cs

irvinesunday: Quite true. Thanks

andrueastman (approved):
Thanks for this!

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
baywet: I believe it should in fact be 
```CSharp
if (context.Settings.EnableDefaultValueForOdataTypeProperty && (isBaseTypeAbstractNonEntity || context.Model.IsBaseTypeReferencedAsTypeInModel(structuredType.BaseType)))
```
since you always want the abstraction aspect to be considered, not just the setting itself. Does that make sense?

baywet (changes requested):
thanks for taking the changes, one last thing

irvinesunday:
> thanks for taking the changes, one last thing

From the issue description written by @andrueastman, I understood that we should check whether the type derives from an abstract type alongside the other 2 listed checks, else, we generate the `default` value by default (as is the current current case). Is this not the expectation?

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs
andrueastman: In my head, we are trying to a solve a problem specific to graph as the existence of the property by default should not really be a problem for an OData API. 

Currently, the setting therefore creates the possibility of disabling this "custom" behavior, by enabling all models to have the property set by default, otherwise having them only set when type disambiguation is really needed. 

The proposed change would have the setting either have the property set only when type disambiguation is really needed or not at all. Which still works as we can use the setting to get what we want. 

I think the proposal by @irvinesunday makes sense with a possible renaming of the setting to `EnableOdataTypePropertyOnlyForTypeDisambiguation` so that it makes it clearer in the context.

baywet: aaah ok, thanks for clarifying here, I think I missed some of the intent. So we're still going to always add the odata type by default unless the disambiguation is on AND we match conditions when not to include it to avoid ambiguous situations.
On top of the setting rename, I'd like to suggest setting it to false by default.

The condition should then be something like

```CSharp
if (!context.Settings.EnableDefaultValueForOdataTypeProperty || (context.Settings.EnableDefaultValueForOdataTypeProperty && (isBaseTypeAbstractNonEntity || context.Model.IsBaseTypeReferencedAsTypeInModel(structuredType.BaseType))))
```
makes sense?

andrueastman: Agreed!

irvinesunday: Renamed the convert setting to `EnableTypeDisambiguationForDefaultValueOfOdataTypeProperty` and simplified the condition to:
```
if (!context.Settings.EnableTypeDisambiguationForDefaultValueOfOdataTypeProperty ||
    isBaseTypeAbstractNonEntity ||                  
    context.Model.IsBaseTypeReferencedAsTypeInModel(structuredType.BaseType))
```

baywet: right boolean algebra simplifications, this takes me back :)

PR #308, opened 2022-11-11
Add support for paths with alternate keys

### LINKED ISSUE #120: Add support for alternate keys
Models containing alternate keys cannot easily be converted to OpenAPI. Corresponding paths provided by a custom path provider lead to an unhandled exception in parameter creation.

### Assemblies affected

All versions of Microsoft.OpenApi.OData.Reader.dll

### Steps to reproduce

See the following modifications of the single key test. It reflects a case for an alternate key path http://host/service/Customer(AlternateId='A1245') which was added by a custom `ODataPathProvider` handed over in the `OpenAPIConvertSettings`. The same can be achieved for multi-part alternate keys.

```diff
[Theory]
[InlineData(true)]
[InlineData(false)]
public void CreateKeyParametersForSingleKeyWorks(bool prefix)
{
    // Arrange
    EdmModel model = new EdmModel();
    EdmEntityType customer = new EdmEntityType("NS", "Customer");
    customer.AddKeys(customer.AddStructuralProperty("Id", EdmPrimitiveTypeKind.String));

+    var alternateId = customer.AddStructuralProperty("AlternateId", EdmPrimitiveTypeKind.String);
+    model.AddAlternateKeyAnnotation(customer, new System.Collections.Generic.Dictionary<string, IEdmProperty> { { "AlternateId", alternateId } });

    model.AddElement(customer);
    OpenApiConvertSettings setting = new OpenApiConvertSettings
    {
        PrefixEntityTypeNameBeforeKey = prefix
    };
    ODataContext context = new ODataContext(model, setting);
    ODataKeySegment keySegment = new ODataKeySegment(customer);
    // Act
+    var parameters = context.CreateKeyParameters(keySegment, new System.Collections.Generic.Dictionary<string, string> { { "AlternateId", "alternateId" } });
-    var parameters = context.CreateKeyParameters(keySegment);
...
```

### Expected result

No exception without considering alternate keys.

### Actual result

KeyNotFoundException in OpenApiParameterGenerator.CreateKeyParameters

### Additional detail

Given the mere example `http://host/service/Employees(ID='A1245')` from [OData Version 4.01. Part 2: URL Conventions](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_AlternateKeys), without a model it is unclear whether one should rely on the (primary) key or an alternate key. Since the model is known for the generator, not only the primary key but also the alternative keys could be taken into account when mapping the parameter names. Definitely it should work fault tolerant with any other PathProvider then the default ODataPathProvider.

millicentachieng:
Fixes #120 

This PR:
- Adds new paths for entities that have alternate keys
- Renames the summary and OperationId of operations to include key name
- Add a setting to enable the generation of paths with alternate keys

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -50,7 +50,7 @@ namespace Microsoft.OpenApi.OData.Common
                 return input;
             }
 
-            char first = Char.ToUpper(input[0]);
+            char first = char.ToUpper(input[0]);
             return first + input.Substring(1);
         }
 
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataKeySegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataKeySegment.cs
@@ -60,7 +60,7 @@ namespace Microsoft.OpenApi.OData.Edm
             {
                 IList<string> keys = IsAlternateKey ? 
                     KeyMappings.Values.ToList() : 
-                    EntityType.Key().Select(x => x.Name).ToList();
+                    EntityType.Key().Select(static x => x.Name).ToList();
 
                 return string.Join(",", keys);
             }
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataKeySegment.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataKeySegment.cs
@@ -38,6 +38,11 @@ namespace Microsoft.OpenApi.OData.Edm
         }
 
         /// <summary>
+        /// Is true if key segment is alternate key
+        /// </summary>
+        public bool IsAlternateKey { get; set; } = false;
+
+        /// <summary>
         /// Gets the key/template mappings.
         /// </summary>
         public IDictionary<string, string> KeyMappings { get; }
@@ -53,13 +58,11 @@ namespace Microsoft.OpenApi.OData.Edm
         {
             get
             {
-                IList<string> keys = new List<string>();
-                foreach (var key in EntityType.Key())
-                {
-                    keys.Add(key.Name);
-                }
+                IList<string> keys = IsAlternateKey ? 
+                    KeyMappings.Values.ToList() : 
+                    EntityType.Key().Select(x => x.Name).ToList();
 
-                return String.Join(",", keys);
+                return string.Join(",", keys);
             }
         }
 
@@ -77,7 +80,7 @@ namespace Microsoft.OpenApi.OData.Edm
             // Use the output key/template mapping
             if (KeyMappings != null)
             {
-                if (KeyMappings.Count == 1)
+                if (KeyMappings.Count == 1 && !IsAlternateKey)
                 {
                     var key = KeyMappings.First();
                     return $"{{{key.Value}}}";
@@ -114,7 +117,7 @@ namespace Microsoft.OpenApi.OData.Edm
                     keyStrings.Add($"{keyProperty.Name}={quote}{{{name}}}{quote}");
                 }
 
-                return String.Join(",", keyStrings);
+                return string.Join(",", keyStrings);
             }
         }
 
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
@@ -45,7 +45,7 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <param name="settings">The conversion settings.</param>
         /// <returns>The collection of built <see cref="ODataPath"/>.</returns>
         public virtual IEnumerable<ODataPath> GetPaths(IEdmModel model, OpenApiConvertSettings settings)
-       {
+        {
            if (model == null || model.EntityContainer == null)
            {
                return Enumerable.Empty<ODataPath>();
@@ -92,7 +92,7 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <param name="model">The Edm model.</param>
         protected virtual void Initialize(IEdmModel model)
         {
-            Debug.Assert(model != null);
+            Utils.CheckArgumentNull(model, nameof(model));
 
             _model = model;
             _allNavigationSources = model.LoadAllNavigationSources();
@@ -123,7 +123,7 @@ namespace Microsoft.OpenApi.OData.Edm
 
         private void AppendPath(ODataPath path)
         {
-            Debug.Assert(path != null);
+            Utils.CheckArgumentNull(path, nameof(path));
 
             ODataPathKind kind = path.Kind;
             switch(kind)
@@ -176,7 +176,8 @@ namespace Microsoft.OpenApi.OData.Edm
         /// <param name="convertSettings">The settings for the current conversion.</param>
         private void RetrieveNavigationSourcePaths(IEdmNavigationSource navigationSource, OpenApiConvertSettings convertSettings)
         {
-            Debug.Assert(navigationSource != null);
+            Utils.CheckArgumentNull(navigationSource, nameof(navigationSource));
+            Utils.CheckArgumentNull(convertSettings, nameof(convertSettings));
 
             // navigation source itself
             ODataPath path = new(new ODataNavigationSourceSegment(navigationSource));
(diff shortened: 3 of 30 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -50,7 +50,7 @@ public static string UpperFirstChar(string input)
                 return input;
             }
 
-            char first = Char.ToUpper(input[0]);
```
xuzhg, line 53: No need to change these to unfocus the main priority in one PR.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -516,14 +523,44 @@ private void CreateRefPath(ODataPath currentPath)
...
+        /// <param name="entityType">The entityType with alternate keys</param>
+        private void CreateAlternateKeyPath(ODataPath currentPath, IEdmEntityType entityType)
+        {
+            Utils.CheckArgumentNull(currentPath, nameof(currentPath));
+            Utils.CheckArgumentNull(entityType, nameof(entityType));
+
+            IEnumerable<IDictionary<string, IEdmProperty>> alternateKeys = _model.GetAlternateKeysAnnotation(entityType);
+            if (alternateKeys.Count() > 0)
+            {
+                foreach (var keyDict in alternateKeys)
```
xuzhg, line 550: calling .Count() and do foreach will perform double evaluation for the IEnumerable< T >. Maybe consider to call .ToList() or .ToArray() first?

test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
```diff
@@ -617,6 +618,89 @@ public void GetPathsWithStreamPropertyAndWithEntityHasStreamWorks(bool hasStream
             }
         }
 
+        [Fact]
+        public void GetPathsWithAlternateKeyParametersWorks()
+        {
+            string alternateKeyProperty =
+@"<Property Name=""SSN"" Type=""Edm.String""/>
+    <Annotation Term=""OData.Community.Keys.V1.AlternateKeys"">
```
xuzhg, line 626: ODL supports to use OData.Core.Vocabulary alternateKeys. And that's the recommended solution.

Consider to enable this.

xuzhg (dismissed):

test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
millicentachieng: For some reason graph metadata uses `OData.Community.Keys.V1.AlternateKey`. I'm changing the test to use `Org.OData.Core.V1.AlternateKey`.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataKeySegment.cs
```diff
@@ -53,13 +58,11 @@ public override string Identifier
...
             get
             {
-                IList<string> keys = new List<string>();
-                foreach (var key in EntityType.Key())
-                {
-                    keys.Add(key.Name);
-                }
+                IList<string> keys = IsAlternateKey ? 
+                    KeyMappings.Values.ToList() : 
+                    EntityType.Key().Select(x => x.Name).ToList();
```
baywet, line 63:
```suggestion
                    EntityType.Key().Select(static x => x.Name).ToList();
```

src/Microsoft.OpenApi.OData.Reader/Operation/EntityDeleteOperationHandler.cs
```diff
@@ -36,17 +36,19 @@ protected override void Initialize(ODataContext context, ODataPath path)
...
+            string placeHolder = $"Delete entity from {EntitySet.Name} by key ({keySegment.Identifier})";
             operation.Summary = _deleteRestrictions?.Description ?? placeHolder;
             operation.Description = _deleteRestrictions?.LongDescription;
 
             // OperationId
             if (Context.Settings.EnableOperationId)
             {
                 string typeName = entityType.Name;
-                operation.OperationId = EntitySet.Name + "." + typeName + ".Delete" + Utils.UpperFirstChar(typeName);
+                string keyName = string.Join("", keySegment.Identifier.Split(',').Select(x => Utils.UpperFirstChar(x)));
```
baywet, line 50:
```suggestion
                string keyName = string.Join("", keySegment.Identifier.Split(',').Select(static x => Utils.UpperFirstChar(x)));
```

src/Microsoft.OpenApi.OData.Reader/Operation/EntityGetOperationHandler.cs
```diff
@@ -37,17 +37,19 @@ protected override void Initialize(ODataContext context, ODataPath path)
...
             operation.Summary = _readRestrictions?.ReadByKeyRestrictions?.Description ?? placeHolder;
             operation.Description = _readRestrictions?.ReadByKeyRestrictions?.LongDescription ?? Context.Model.GetDescriptionAnnotation(entityType);
 
             // OperationId
             if (Context.Settings.EnableOperationId)
-            {
+            { 
                 string typeName = entityType.Name;
-                operation.OperationId = EntitySet.Name + "." + typeName + ".Get" + Utils.UpperFirstChar(typeName);
+                string keyName = string.Join("", keySegment.Identifier.Split(',').Select(x => Utils.UpperFirstChar(x)));
```
baywet, line 51:
```suggestion
                string keyName = string.Join("", keySegment.Identifier.Split(',').Select(static x => Utils.UpperFirstChar(x)));
```

src/Microsoft.OpenApi.OData.Reader/Operation/EntityUpdateOperationHandler.cs
```diff
@@ -32,17 +32,19 @@ protected override void Initialize(ODataContext context, ODataPath path)
...
+            string placeHolder = $"Update entity in {EntitySet.Name} by key ({keySegment.Identifier})";
             operation.Summary = _updateRestrictions?.Description ?? placeHolder;
             operation.Description = _updateRestrictions?.LongDescription;
 
             // OperationId
             if (Context.Settings.EnableOperationId)
             {
                 string typeName = entityType.Name;
-                operation.OperationId = EntitySet.Name + "." + typeName + ".Update" + Utils.UpperFirstChar(typeName);
+                string keyName = string.Join("", keySegment.Identifier.Split(',').Select(x => Utils.UpperFirstChar(x)));
```
baywet, line 46:
```suggestion
                string keyName = string.Join("", keySegment.Identifier.Split(',').Select(static x => Utils.UpperFirstChar(x)));
```

baywet (changes requested):
I also believe we should introduce a setting for that new behavior

millicentachieng:
> I also believe we should introduce a setting for that new behavior

I had a discussion with @darrelmiller regarding this, and we thought that if there are alternate keys defined in the metadata then we ought to generate paths that utilize them as well. 

Do you think the presence of these paths is going to affect any downstream tools? CC @peombwa

baywet:
It might increase significantly the size of the SDKs, especially for sets that are fairly "close to the root" of the API paths tree since the logic might go ahead and expand all the paths we already had under the main key.
Additionally, the "custom" has been to introduce settings as a way for consumers to control the behaviour of the conversion every time we've introduced changes that add additional paths to the resulting description. This way, should somebody complain about the new paths and not want them, they have a workaround.

baywet:
Thanks, that's what I suspected from the code but I wasn't sure whether it was a final decision or a stopgap.

baywet (approved):
thanks for making the changes!

peombwa:
> > I also believe we should introduce a setting for that new behavior
> 
> I had a discussion with @darrelmiller regarding this, and we thought that if there are alternate keys defined in the metadata then we ought to generate paths that utilize them as well.
> 
> Do you think the presence of these paths is going to affect any downstream tools? CC @peombwa

As @baywet said above, we may see a significant growth in paths. However, having the setting to control how paths with alternate keys are generated is enough to warrant any concerns with downstream tools.

peombwa (approved):

PR #317, opened 2022-12-06
Resolve references when converting EdmModel to OpenApiDocument

### LINKED ISSUE #171: OpenAPIReference objects are not created correctly
When creating OpenAPI model objects that are just references to another model object they need to be marked as `unresolved` so that reference resolution can replace them with the resolved object.

When an object is created and assigned an OpenApiReference like this, 
```csharp
 OpenApiSchema derivedTypeSchema = new()
				{
                    Reference = new OpenApiReference
                    {
                        Type = ReferenceType.Schema,
                        Id = derivedType.FullName()
                    }
                };
```

there needs to be an assignment to the UnresolvedReference property.

```csharp
 OpenApiSchema derivedTypeSchema = new()
				{
                    UnresolvedReference = true,
                    Reference = new OpenApiReference
                    {
                        Type = ReferenceType.Schema,
                        Id = derivedType.FullName()
                    }
                };
```

Without this flag set, client code will assume that the object is just an empty model object.  The presence of an OpenApiReference object is not sufficient to identify a reference as model objects that are targeted by the reference will also have the OpenApiReference set.

millicentachieng:
Fixes #171

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/EdmModelOpenApiExtensions.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/EdmModelOpenApiExtensions.cs
@@ -3,7 +3,6 @@
 //  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // ------------------------------------------------------------
 
-using System.Collections.Generic;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Edm.Validation;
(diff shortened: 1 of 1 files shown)


millicentachieng:
This is not the solution we are looking for. More details [here](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1144#issuecomment-1418569316).

Issue #313, opened 2022-11-28
Missing query parameters for raw count endpoints


baywet:
The raw count endpoints ($count path item) are missing query parameters descriptions from the parent endpoint.

The service supports `https://graph.microsoft.com/v1.0/me/messages/$count?$filter=isDraft eq false` but the conversion process is missing the query parameters from messages on the $count path item.

### Assemblies affected

1.2.0-preview8

### Expected result

$count endpoints to have query parameters from the parent set/collection (filter, search and custom parameters)
http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_AddressingtheCountofaCollection

calebkiage:
@baywet, the headers are also missing

irvinesunday:
> @baywet, the headers e.g. `Consistency-Level` are also missing

@calebkiage this support was added in this PR: https://github.com/microsoft/OpenAPI.NET.OData/pull/305

PR #320, opened 2022-12-14
Adds OData query parameters to `$count` endpoints

irvinesunday:
This PR:

- Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/313
- Appends OData query parameters to `$count` endpoints.
- Updates integration test files.
- Updates release notes.


src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
```diff
@@ -36,6 +39,16 @@ protected override void Initialize(ODataContext context, ODataPath path)
             int count = path.Segments.Count;
             if(count >= SecondLastSegmentIndex)
                 LastSecondSegment = path.Segments.ElementAt(count - SecondLastSegmentIndex);
+
+            if (LastSecondSegment is ODataNavigationSourceSegment sourceSegment)
+            {
+                annotatable = sourceSegment.NavigationSource as IEdmEntitySet;
+                annotatable ??= sourceSegment.NavigationSource as IEdmSingleton;
```
baywet, line 46:
```suggestion
                annotatable = sourceSegment.NavigationSource as IEdmEntitySet ?? sourceSegment.NavigationSource as IEdmSingleton;
```

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
```diff
@@ -75,19 +88,85 @@ protected override void SetResponses(OpenApiOperation operation)
...
+                    operation.Parameters.Add(parameter);
+                }
             }
-            else if (LastSecondSegment is ODataNavigationPropertySegment navigationPropertySegment)
+            else
             {
-                annotatable = navigationPropertySegment.NavigationProperty;
+                // EntitySet or Collection-Valued Navigation Property
+
+                OpenApiParameter parameter = Context.CreateTop(annotatable);
```
baywet, line 115: > The /$count path suffix identifies the integer count of records in the collection and SHOULD NOT be combined with the system query options  [$top](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptionstopandskip), [$skip](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptionstopandskip), [$orderby](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptionorderby), [$expand](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptionexpand), and [$format](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptionformat). The count MUST NOT be affected by $top, $skip, $orderby, or $expand.

Please remove skip, top, search, expand, count, select, orderby

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
```diff
@@ -75,19 +88,85 @@ protected override void SetResponses(OpenApiOperation operation)
             base.SetResponses(operation);
         }
 
-        protected override void AppendCustomParameters(OpenApiOperation operation)
+        /// <inheritdoc/>
+        protected override void SetParameters(OpenApiOperation operation)
         {
-            IEdmVocabularyAnnotatable annotatable = null;            
-            if (LastSecondSegment is ODataNavigationSourceSegment sourceSegment)
+            base.SetParameters(operation);
```
baywet, line 94: this should also support custom defined request headers like Consistency Level defined on the parent.

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
irvinesunday: We already add custom header here: https://github.com/microsoft/OpenAPI.NET.OData/blob/e7bad1197d1469d2cbffaebdefae470fb7cc7043/src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs#L103-L106

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
irvinesunday: That means we only create `$filter` and `$search` and only if the parent endpoint is a collection-valued nav. prop. or an entity set?

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
irvinesunday: Getting an error here: _Operator '??' cannot be applied to operands of type IEdmEntitySet and IEdmSingleton_. Reverting to original code.

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
baywet: correct, I believe the raw count path segment will only ever get added after those anyway with the code we have today.

src/Microsoft.OpenApi.OData.Reader/Operation/DollarCountGetOperationHandler.cs
baywet: could we use a common parent interface instead? like IAnnotable or something similar?

irvinesunday: I have opted to do away with the `IEdmSingleton` cast. We are not using it anyway

baywet (approved):

Issue #324, opened 2022-12-22
OData cast paths should follow the same naming convention for `tags` and `operationIds` as non-OData cast paths


peombwa:
To support accurate slicing of an OpenAPI document, OData cast paths should follow the same naming convention for `tags` and `operationIds` as their non-OData cast paths counterparts. An `As<TypeName>` should be added at the end of an operationId.

### Assemblies affected
[v1.1.0](https://github.com/microsoft/OpenAPI.NET.OData/releases/tag/v1.1.0)

### Actual result

| Path | Tags | OperationId |
| - | - | - |
| `GET /directoryRoles/{directoryRole-id}/members` |  `directoryRoles.directoryObject` | `directoryRoles.ListMembers` |
| **`GET /directoryRoles/{directoryRole-id}/members/microsoft.graph.application`** |  `directoryObject.application` | `Get.microsoft.graph.directoryObject.Items.As.microsoft.graph.application-ced0` |

### Expected result

| Path | Tags | OperationId |
| - | - | - |
| `GET /directoryRoles/{directoryRole-id}/members` |  `directoryRoles.directoryObject` | `directoryRoles.ListMembers` |
| ~~`GET /directoryRoles/{directoryRole-id}/members/microsoft.graph.application`~~ |  ~~`directoryObject.application`~~ | ~~`Get.microsoft.graph.directoryObject.Items.As.microsoft.graph.application-ced0`~~ |
| **`GET /directoryRoles/{directoryRole-id}/members/microsoft.graph.application`** |  `directoryRoles.directoryObject` | `directoryRoles.ListMembers.AsApplication` |

irvinesunday:
How about operation ids of `$count` segments appended to OData cast paths? Example:

```YAML
 /directoryRoles/{directoryRole-id}/members/microsoft.graph.device/$count':
  description: Provides operations to count the resources in the collection.
  get:
    summary: Get the number of the resource
    operationId: Get.Count.microsoft.graph.device-e712
```
Would the operation id: `Get.Count.microsoft.graph.device-e712` also need to be updated as `Get.Count.AsDevice-e712`?

irvinesunday:
Currently, we don't generate tag names for `$count` and complex property paths. 
What tag naming convention do you think will be most appropriate for:
1. `$count` paths. Example is the one mentioned above
2. Complex property paths. Example:
```YAML
'/admin/serviceAnnouncement/issues/{serviceHealthIssue-id}/posts':
    get:
[...]

PR #338, opened 2023-02-01
Resolves `operationId` and `tag` names for OData cast paths

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/324

This PR:
- Renames the `operationId` and `tag` names for OData cast paths to conform with other no non-OData cast paths.

**NB:** The PR is currently in draft because:
- The relevant Integration and unit tests haven't been updated yet. I will need to lock in any requested changes before updating these.
- Complex properties and `$count` paths currently don't have tags generated for them. We need to agree on a format before these can be updated into this fix. The discussion can be found [here](https://github.com/microsoft/OpenAPI.NET.OData/issues/324#issuecomment-1411686480).

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OData.Edm;
 using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.OData.Edm;
 using Microsoft.OpenApi.OData.Vocabulary.Capabilities;
 
 namespace Microsoft.OpenApi.OData.Common
@@ -22,7 +23,7 @@ namespace Microsoft.OpenApi.OData.Common
         {
             Utils.CheckArgumentNull(structuredType, nameof(structuredType));
             Utils.CheckArgumentNull(edmModel, nameof(edmModel));
-            if(structuredType is not IEdmSchemaElement schemaElement) throw new ArgumentException("The type is not a schema element.", nameof(structuredType));
+            if (structuredType is not IEdmSchemaElement schemaElement) throw new ArgumentException("The type is not a schema element.", nameof(structuredType));
 
             IEnumerable<IEdmSchemaElement> derivedTypes = edmModel.FindAllDerivedTypes(structuredType).OfType<IEdmSchemaElement>();
 
@@ -32,12 +33,12 @@ namespace Microsoft.OpenApi.OData.Common
             }
 
             OpenApiSchema schema = new()
-			{
+            {
                 OneOf = new List<OpenApiSchema>()
             };
 
             OpenApiSchema baseTypeSchema = new()
-			{
+            {
                 UnresolvedReference = true,
                 Reference = new OpenApiReference
                 {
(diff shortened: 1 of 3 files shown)


irvinesunday:
@baywet and @peombwa I might need some much-need eyes on this draft PR before I proceed to add the tests :)

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -83,7 +84,91 @@ internal static OpenApiSchema GetDerivedTypesReferenceSchema(IEdmStructuredType
...
+        }
+
+        /// <summary>
+        /// Generates the operation id for a navigation property path.
+        /// </summary>
+        /// <param name="path">The target <see cref="ODataPath"/>.</param>
+        /// <param name="navigationSource">The <see cref="IEdmNavigationSource"/> of the target path.</param>
+        /// <param name="prefix">Identifier indicating whether it is a collection-valued non-indexed navigation property.</param>
+        /// <returns>The operation id name.</returns>
+        internal static string GenerateNavigationPropertyPathOperationId(ODataPath path, IEdmNavigationSource navigationSource, string prefix = null)
```
peombwa, line 97: Can we standardize the naming convention of operationIds to what we have in DevX API - https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/dev/OpenAPIService/PowershellFormatter.cs?
For example:
- Remove hash suffix values from operationIds of function paths.
- Add '_' to separate verb (action) in an operationId. This typically the last `.` or second `.` for OData cast paths.
- PUT operations should have Set as the verb in an operationId `{xxx}_Set{yyy}`.

In PowerShell, we use operationIds to form command names.

peombwa: See my comments at https://github.com/microsoft/OpenAPI.NET.OData/issues/324#issuecomment-1419518641 on operationIds and tags for count and complex property paths.

irvinesunday: > Can we standardize the naming convention of operationIds to what we have in DevX API - https://github.com/microsoftgraph/microsoft-graph-devx-api/blob/dev/OpenAPIService/PowershellFormatter.cs? For example:
> 
> * Remove hash suffix values from operationIds of function paths.
> * Add '_' to separate verb (action) in an operationId. This typically the last `.` or second `.` for OData cast paths.
> * PUT operations should have Set as the verb in an operationId `{xxx}_Set{yyy}`.
> 
> In PowerShell, we use operationIds to form command names.

There are other clients who rely on the operation ids, not sure it will be a good idea to handle that in this library. See this [issue](https://github.com/microsoftgraph/msgraph-metadata/issues/289).

peombwa: They are also using AutoREST in the referenced issue - https://github.com/microsoftgraph/msgraph-metadata/issues/289. Standardizing the operationIds here and ensuring uniqueness will have the net effect of fixing their issue.

To unblock this change, I've created an issue at https://github.com/microsoft/OpenAPI.NET.OData/issues/361 for us to discuss this further.

peombwa (dismissed):

peombwa (approved):

Issue #343, opened 2023-02-21
Duplicate paths causes failed transformed metadata


andrueastman:
At the moment, Hidi fails to generate the latest beta openApi file for Microsoft graph with the following error

```shell
Tool 'microsoft.openapi.hidi' (version '1.2.2') was successfully installed.
crit: Microsoft.OpenApi.Hidi.OpenApiService[0]
      Could not transform the document, reason: An item with the same key has already been added. Key: /administrativeUnits/microsoft.graph.delta()
```
Reference https://dev.azure.com/microsoftgraph/Graph%20Developer%20Experiences/_build/results?buildId=106293&view=logs&j=4c81e8e4-6eb9-5def-7261-b44ad0fc1b7d&t=163839c6-f69e-5b5d-39e2-5b38e97d3fe3

Digging further into this it looks like the latest metadata has introduced the delta function to the directoryObjectt ype which is the base type of `administrativeUnit`

```xml
      <Function Name="delta" IsBound="true">
        <Parameter Name="bindingParameter" Type="Collection(graph.directoryObject)" />
        <ReturnType Type="Collection(graph.directoryObject)" />
      </Function>
```
Due to the check done here to try to apply bound functions on all the derived types of the object, two paths are generated due to the already existing function on the base type and existing function on the actual type. 

https://github.com/microsoft/OpenAPI.NET.OData/blob/6e6442946e21c597078de3b25ab85290d6d7db63/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L668

This causes the generation to fail due to the duplicated keys when trying to build the dictionary of paths at 
https://github.com/microsoft/OpenAPI.NET.OData/blob/master/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiPathItemGenerator.cs#L58

Notes
- Is it incorrect to have a bound function on a base type redeclared on the child type? (This would mean that the metadata is incorrect)
- We should probably try to use `Dictionary.TryAdd` and maybe log a warning in the event that we find duplicate paths to prevent failures
- Paths generated from derived types could append a suffix to ensure their uniqueness to if the metadata is valid. 

cc @irvinesunday @baywet

baywet:
happy that we came to the same conclusions, [for reference](https://github.com/microsoftgraph/msgraph-metadata/issues/276).
[...]

PR #349, opened 2023-02-23
Checks whether path exists before adding it to the paths dictionary

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/343

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -272,7 +272,6 @@ namespace Microsoft.OpenApi.OData.Common
             return false;
         }
 
-
         /// <summary>
         /// Gets the entity type of the target <paramref name="segment"/>.
         /// </summary>
@@ -297,8 +296,6 @@ namespace Microsoft.OpenApi.OData.Common
                 default:
                     return null;
             }
-
-            return null;
         }
 
         /// <summary>
@@ -366,14 +363,16 @@ namespace Microsoft.OpenApi.OData.Common
                 }
 
                 ODataSegment lastSecondSegment = path.Segments.ElementAt(path.Count - secondLastSegmentIndex);
-                IEdmEntityType boundEntityType = lastSecondSegment.EntityTypeFromPathSegment();
+                IEdmEntityType boundEntityType = lastSecondSegment?.EntityTypeFromPathSegment();
 
                 IEdmEntityType operationEntityType = lastSegment.EntityTypeFromOperationSegment();
-                IEnumerable<IEdmStructuredType> derivedTypes = context.Model.FindAllDerivedTypes(operationEntityType);
+                IEnumerable<IEdmStructuredType> derivedTypes = (operationEntityType != null)
+                    ? context.Model.FindAllDerivedTypes(operationEntityType)
+                    : null;
 
-                if (derivedTypes.Any())
+                if (derivedTypes?.Any() ?? false)
                 {
-                    if (!derivedTypes.Contains(boundEntityType))
+                    if (boundEntityType != null && !derivedTypes.Contains(boundEntityType))
                     {
                         Debug.WriteLine($"Duplicate paths present but entity type of binding parameter '{operationEntityType}' " +
                                         $"is not the base type of the bound entity type '{boundEntityType}'. Path: {pathName}");
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -276,8 +276,8 @@ namespace Microsoft.OpenApi.OData.Common
         /// <summary>
         /// Gets the entity type of the target <paramref name="segment"/>.
         /// </summary>
-        /// <param name="segment">The target <see cref="ODataSegment"/></param>
-        /// <returns>The entity type of the target <paramref name="segment"/></returns>
+        /// <param name="segment">The target <see cref="ODataSegment"/>.</param>
+        /// <returns>The entity type of the target <paramref name="segment"/>.</returns>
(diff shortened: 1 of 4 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,131 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
             return false;
         }
+
+
+        /// <summary>
+        /// Gets the entity type of the target <paramref name="segment"/>.
+        /// </summary>
+        /// <param name="segment">The target <see cref="ODataSegment"/></param>
+        /// <returns>The entity type of the target <paramref name="segment"/></returns>
+        internal static IEdmEntityType EntityTypeFromPathSegment(this ODataSegment segment)
```
andrueastman, line 281:
```suggestion
        private static IEdmEntityType EntityTypeFromPathSegment(this ODataSegment segment)
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,132 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                }                    
+            }
+            else if (segment is ODataKeySegment keySegment)
+            {
+                return keySegment.EntityType;
+            }
+            else if (segment is ODataOperationSegment)
+            {
+                return segment.EntityTypeFromOperationSegment();
+            }
```
andrueastman, line 307: A switch statement would be nice to use here.
```suggestion
            switch (segment)
            {
                case ODataNavigationPropertySegment navPropSegment:
                    return navPropSegment.EntityType;
                case ODataNavigationSourceSegment navSourceSegment when navSourceSegment.NavigationSource is IEdmEntitySet entitySet:
                    return entitySet.EntityType();
                case ODataNavigationSourceSegment navSourceSegment when navSourceSegment.NavigationSource is IEdmSingleton singleton:
                    return singleton.EntityType();
                case ODataKeySegment keySegment:
                    return keySegment.EntityType;
                case ODataOperationSegment:
                    return segment.EntityTypeFromOperationSegment();
                default:
                    return null;
            }
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,132 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                if (path.Count < secondLastSegmentIndex)
+                {
+                    Debug.WriteLine($"Invalid path. Operation not bound to any entity. Path: {pathName}");
+                    return false;
+                }
+
+                ODataSegment lastSecondSegment = path.Segments.ElementAt(path.Count - secondLastSegmentIndex);
+                IEdmEntityType boundEntityType = lastSecondSegment.EntityTypeFromPathSegment();
+
+                IEdmEntityType operationEntityType = lastSegment.EntityTypeFromOperationSegment();
```
andrueastman, line 379: As these two functions can return null, we should probably check for null returns before using the returned values later on.

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: This extension method will def. be reused outside this class.

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,123 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+        /// </summary>
+        /// <param name="segment">The target <see cref="ODataOperationSegment"/>.</param>
+        /// <returns>The entity type of the target <paramref name="segment"/>.</returns>
+        private static IEdmEntityType EntityTypeFromOperationSegment(this ODataSegment segment)
+        {
+            CheckArgumentNull(segment, nameof(segment));
+
+            if (segment is ODataOperationSegment operationSegment)
+            {
+                IEdmOperationParameter bindingParameter = operationSegment.Operation.Parameters.First();
```
baywet, line 312:
```suggestion
            if (segment is ODataOperationSegment operationSegment &&
            operationSegment.Operation.Parameters.FirstOrDefault() is IEdmOperationParameter bindingParameter)
            {
```

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,123 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                if (path.LastSegment is not ODataOperationSegment lastSegment)
+                {
+                    Debug.WriteLine("Duplicate path: " + pathName);
+                    return false;
+                }
+
+                int secondLastSegmentIndex = 2;
+                if (path.Count < secondLastSegmentIndex)
+                {
+                    Debug.WriteLine($"Invalid path. Operation not bound to any entity. Path: {pathName}");
```
baywet, line 361: we could have the same issue with an operation bound to a collection, and have a collection at the root. E.g. have one activate method on directory object collections, and and another activate on users collection. All of a sudden /users/activate becomes a problem.

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: _/users/activate_ will not fall within that conditional statement as the no. of segments will be equal to 2

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: Resolved

baywet (dismissed):

baywet (approved):

Issue #348, opened 2023-02-23
Adds a setting to strip namespace prefix from path for OData methods


baywet:
right now we're projecting
/me/microsoft.graph.sendMail

with that setting, set to microsoft.graph we should be able to project instead

/me/sendMail

If a function/action is defined in a sub namespace like

/.../microsoft.graph.callRecords.activate

it should keep the subnamespace like so

/.../callRecords.activate

irvinesunday:
Namespaces like `microsoft.graph.callRecords`, `microsoft.graph.security` and `microsoft.graph` are fully qualified namespaces. Why would we only want to use the last segments e.g. for the 1st two namespaces?

andrueastman:
> Namespaces like `microsoft.graph.callRecords`, `microsoft.graph.security` and `microsoft.graph` are fully qualified namespaces. Why would we only want to use the last segments e.g. for the 1st two namespaces?
> 
> We currently use the setting `EnableUnqualifiedCall` to generate operation names without their namespaces.
> 
> https://github.com/microsoft/OpenAPI.NET.OData/blob/a8d815a19f8feae41c55fc66779d25d26ef7eb78/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs#L38

Does that mean changing the setting would change the paths currently generated from 

```
/me/microsoft.graph.sendMail
```
to 
```
/me/sendMail
```

irvinesunday:
> > Namespaces like `microsoft.graph.callRecords`, `microsoft.graph.security` and `microsoft.graph` are fully qualified namespaces. Why would we only want to use the last segments e.g. for the 1st two namespaces?
> > We currently use the setting `EnableUnqualifiedCall` to generate operation names without their namespaces.
> > https://github.com/microsoft/OpenAPI.NET.OData/blob/a8d815a19f8feae41c55fc66779d25d26ef7eb78/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs#L38
> 
> Does that mean changing the setting would change the paths currently generated from
> 
> ```
> /me/microsoft.graph.sendMail
> ```
> 
> to
> 
> ```
> /me/sendMail
> ```

Correct

irvinesunday:
We currently use the fully qualified namespace names in operations and type cast segments. Are we anticipating this setting to simplify the naming of both segments?

baywet:
[...]

PR #350, opened 2023-02-24
Strips namespace prefix from operation segments and aliases type cast segments

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/348

This PR:
- Removes the base namespace identifier from operation segments.
- To remove the namespace to make unqualified calls, set the convert setting `EnableUnqualifiedCall=true` and set the `BaseNamespace` value to the base namespace value that needs to be stripped.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -288,7 +288,7 @@ namespace Microsoft.OpenApi.OData.Common
                 return value;
             }
 
-            string defaultNamespace = settings.DefaultNamespace.EndsWith(".")
+            string defaultNamespace = settings.DefaultNamespace.EndsWith(".", StringComparison.OrdinalIgnoreCase)
                 ? settings.DefaultNamespace
                 : settings.DefaultNamespace + ".";
 
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -286,7 +286,7 @@ namespace Microsoft.OpenApi.OData.Common
             // Trim trailing '.' for uniformity
             prefix = prefix.TrimEnd('.');
 
-            return value.StartsWith(prefix)
+            return value.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)
                 ? value.Substring(prefix.Length).TrimStart('.')
                 : value;
         }
--- a/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
@@ -272,27 +272,27 @@ namespace Microsoft.OpenApi.OData.Common
             return false;
         }
 
-        /// <summary>
-        /// Removes the default namespace from a string value.
-        /// </summary>
-        /// <param name="value">The target string value.</param>
-        /// <param name="settings">Convert settings.</param>
-        /// <returns>The string value with the namespace removed.</returns>
-        internal static string RemoveDefaultNamespace(this string value, OpenApiConvertSettings settings)
-        {
-            CheckArgumentNullOrEmpty(value, nameof(value));
-            CheckArgumentNull(settings, nameof(settings));
+        ///// <summary>
+        ///// Removes the default namespace from a string value.
+        ///// </summary>
+        ///// <param name="value">The target string value.</param>
+        ///// <param name="settings">Convert settings.</param>
+        ///// <returns>The string value with the namespace removed.</returns>
+        //internal static string RemoveDefaultNamespace(this string value, OpenApiConvertSettings settings)
+        //{
+        //    CheckArgumentNullOrEmpty(value, nameof(value));
+        //    CheckArgumentNull(settings, nameof(settings));
 
-            if (string.IsNullOrEmpty(settings.DefaultNamespace))
-            {
-                return value;
-            }
+        //    if (string.IsNullOrEmpty(settings.DefaultNamespace))
+        //    {
+        //        return value;
+        //    }
 
-            string defaultNamespace = settings.DefaultNamespace.EndsWith(".", StringComparison.OrdinalIgnoreCase)
-                ? settings.DefaultNamespace
-                : settings.DefaultNamespace + ".";
+        //    string defaultNamespace = settings.DefaultNamespace.EndsWith(".", StringComparison.OrdinalIgnoreCase)
+        //        ? settings.DefaultNamespace
+        //        : settings.DefaultNamespace + ".";
 
-            return value.Replace(defaultNamespace, "");
-        }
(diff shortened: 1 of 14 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,29 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+        {
+            CheckArgumentNullOrEmpty(value, nameof(value));
+            CheckArgumentNull(settings, nameof(settings));
+
+            if (string.IsNullOrEmpty(settings.DefaultNamespace))
+            {
+                return value;
+            }
+
+            string baseNamespace = settings.DefaultNamespace.EndsWith(".")
```
baywet, line 291: pass a string comparison

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,29 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+                ? settings.DefaultNamespace
+                : settings.DefaultNamespace + ".";
+
+            string escapedPattern = Regex.Escape(baseNamespace);
```
baywet, line 295: parsing a regex on each call is going to get really costly. Just using a string.startswith and string.Substring if that's true is probably going to be much faster.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
```diff
@@ -130,14 +131,17 @@ private string FunctionName(IEdmFunction function, OpenApiConvertSettings settin
...
             }
 
-            StringBuilder functionName = new StringBuilder();
+            StringBuilder functionName = new();
+            string selectedName = function.FullName();
+
             if (settings.EnableUnqualifiedCall)
             {
-                functionName.Append(function.Name);
+                selectedName = settings.DefaultNamespace != null ? selectedName.RemoveBaseNamespace(settings) : function.Name;
```
baywet, line 139: move the ternary logic into the method so we don't end up duplicating it all over the place.
check with string.isnullorempty so this doesn't fail if we enable NRT one day.

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
irvinesunday: What's the issue with the current implementation?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
```diff
@@ -112,6 +113,14 @@ public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<
             return parameterNamesMapping;
         }
 
+        private string OperationName(IEdmOperation operation, OpenApiConvertSettings settings)
+        {
+            string selectedName = operation.FullName();
+            return settings.EnableUnqualifiedCall
+                ? settings.DefaultNamespace != null ? selectedName.RemoveDefaultNamespace(settings) : operation.Name
+                : selectedName;
```
baywet, line 121:
```suggestion
            return settings.EnableUnqualifiedCall
                ? selectedName.RemoveDefaultNamespace(settings) : operation.Name;
```

src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -313,6 +313,11 @@ public string PathPrefix
         /// </summary>
         public bool EnableTypeDisambiguationForDefaultValueOfOdataTypeProperty { get; set; } = false;
 
+        /// <summary>
+        /// The default namespace.
+        /// </summary>
+        public string DefaultNamespace{ get; set; }
```
baywet, line 319: see my latest comment on the issue as of how to name the setting

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
baywet: comparison is slower, made a code suggestion, resolving this comment.

src/Microsoft.OpenApi.OData.Reader/Common/Utils.cs
```diff
@@ -270,5 +271,28 @@ internal static void AddCustomAttributesToExtensions(this IDictionary<string, IO
...
+            if (string.IsNullOrEmpty(settings.DefaultNamespace))
+            {
+                return value;
+            }
+
+            string defaultNamespace = settings.DefaultNamespace.EndsWith(".")
+                ? settings.DefaultNamespace
+                : settings.DefaultNamespace + ".";
+
+            return value.Replace(defaultNamespace, "");
```
baywet, line 295:
```suggestion
            return value.StartsWith(defaultNamespace, StringComparison.OrdinalIgnoreCase) ? value.Remove(0, defaultNamespace.Length) : value;
```

irvinesunday: This extension method has been removed in the latest commit.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
irvinesunday: This suggestion is now redundant following the suggestion in the issue discussion.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataOperationSegment.cs
```diff
@@ -112,6 +110,25 @@ public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<
...
+        {
+            if (settings.EnableUnqualifiedCall)
+            {
+                return operation.Name;
+            }
+            else
+            {
+                if (!string.IsNullOrEmpty(settings.NamespacePrefixToStripForInMethodPaths))
+                {
+                    return operation.Namespace.Replace(settings.NamespacePrefixToStripForInMethodPaths, "").TrimStart('.') + "." + operation.Name;
```
baywet, line 123: replace might have side effects if patterns in the naming repeat, startswith + substring instead to avoid such an issue.

baywet (changes requested):

baywet (dismissed):

baywet (dismissed):

baywet:
@irvinesunday since #351 got merged, this is conflicting FYI

irvinesunday:
> @irvinesunday since #351 got merged, this is conflicting FYI

Yes. Resolving the conflicts.

andrueastman (dismissed):

baywet (dismissed):

baywet (approved):

baywet (approved):

Issue #310, opened 2022-11-17
Wrong response status code for Drive upload endpoint


Ndiritu:
The Graph OpenAPI description for `PUT`  `/drives/{drive-id}/items/{driveItem-id}/content` defines a `204` yet the API returns a `201` with either the new/updated drive item resource.

[Docs reference](https://learn.microsoft.com/en-us/graph/api/driveitem-put-content?view=graph-rest-1.0&tabs=http)

peombwa:
This issue also affects `PUT https://graph.microsoft.com/v1.0/me/photo/$value`. Its response code is defined as `204` yet the API returns `200`. See [API reference](https://learn.microsoft.com/en-us/graph/api/profilephoto-update?view=graph-rest-1.0&tabs=http#example-1-update-the-photo-of-the-user).

To resolve the issue, we should ensure that all success status codes in the OpenAPI document are declared as `2XX` when `UseHTTPStatusCodeClass2XX` is set to `true`.

PR #357, opened 2023-03-17
[Fix] Returns response status code 2XX for `PUT` operations of stream properties  when `UseSuccessStatusCodeRange` is enabled

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/310

This PR:
- Returns response status code 2XX for `PUT` operations of stream properties  when `UseSuccessStatusCodeRange` = `true`
- Adds schema reference to the content property of the response object of _content_ stream properties when `UseSuccessStatusCodeRange` = `true`.

Response object of the endpoint: `/drives/{drive-id}/items/{driveItem-id}/content` will be:
image
when `UseSuccessStatusCodeRange` = `true`

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
@@ -27,21 +27,21 @@ namespace Microsoft.OpenApi.OData.Generator
                         Reference = new OpenApiReference
                         {
                             Type = ReferenceType.Response,
-                            Id = "error"
+                            Id = Constants.Error
                         }
                     }
                 },
 
-                { Constants.StatusCode204, new OpenApiResponse { Description = "Success"} },
-                { Constants.StatusCode201, new OpenApiResponse { Description = "Created"} },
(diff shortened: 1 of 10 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiResponseGenerator.cs
```diff
@@ -33,6 +33,8 @@ internal static class OpenApiResponseGenerator
                 },
 
                 { Constants.StatusCode204, new OpenApiResponse { Description = "Success"} },
+                { Constants.StatusCode201, new OpenApiResponse { Description = "Created"} },
+                { Constants.StatusCodeClass2XX, new OpenApiResponse { Description = "Success"} },
```
baywet, line 37: I'm confused, why are we adding the range and the specific codes? shouldn't it be one or the others?

irvinesunday: It's just a dictionary that we can use to extract the response object for the status code range 2XX for scenarios where `addNoContent` == `true` but `schema` == `null`. 
We'll end up with:
image

baywet (approved):

Issue #366, opened 2023-04-04
DELETE methods with no schema have 2XX response.


andrueastman:
Possibly a regression from to https://github.com/microsoft/OpenAPI.NET.OData/pull/357

The Delete methods generated with the latest version of the library now have the 2XX status range as opposes to the 204 code.

example
```yaml
    delete:
      tags:
        - communications.callRecord
      summary: Delete navigation property segments for communications
      operationId: communications.callRecords.sessions.DeleteSegments
      parameters:
        - name: If-Match
          in: header
          description: ETag
          schema:
            type: string
      responses:
        2XX:
          description: Success
        4XX:
          $ref: '#/components/responses/error'
        5XX:
          $ref: '#/components/responses/error'
      x-ms-docs-operation-type: operation
```

This results in Kiota generating return types for these methods for delete requests which may be misleading from a consumer point of view.

irvinesunday:
Should the response type for `Delete` revert to `204` even when `UseSuccessStatusCodeRange` == `true`?

andrueastman:
Yes. I don't think we want to give the information that something could be returned when it won't as this hasn't been the case for DELETE.

PR #367, opened 2023-04-05
[Fix] `DELETE` methods should always return response status code `204`

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/366

This PR:
- Ensures that `delete` methods always return status code `204`. This is regardless of whether `UseSuccessStatusCodeRange` is `true`.
- Adds test to validate the above.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/EntityDeleteOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/EntityDeleteOperationHandler.cs
@@ -81,7 +81,11 @@ namespace Microsoft.OpenApi.OData.Operation
         /// <inheritdoc/>
         protected override void SetResponses(OpenApiOperation operation)
         {
-            operation.AddErrorResponses(Context.Settings, true);
+            // Response for Delete methods should be 204 No Content
+            OpenApiConvertSettings settings = Context.Settings.Clone();
+            settings.UseSuccessStatusCodeRange = false;
+            
+            operation.AddErrorResponses(settings, true);
             base.SetResponses(operation);
         }
 
--- a/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
+++ b/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
@@ -15,20 +15,13 @@
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <PackageId>Microsoft.OpenApi.OData</PackageId>
     <SignAssembly>true</SignAssembly>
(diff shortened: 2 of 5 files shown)


andrueastman (approved):

src/Microsoft.OpenApi.OData.Reader/Operation/EntityDeleteOperationHandler.cs
```diff
@@ -81,7 +81,11 @@ protected override void SetParameters(OpenApiOperation operation)
         /// <inheritdoc/>
         protected override void SetResponses(OpenApiOperation operation)
         {
-            operation.AddErrorResponses(Context.Settings, true);
+            // Response for Delete methods should be 204 No Content
+            OpenApiConvertSettings settings = Context.Settings.Clone();
```
baywet, line 85: I'm a bit late to the party here. But if we could avoid allocs for that kind of thing, it'd be better. Could we not check the operation method/verb in the AddErrorResponses method instead? @irvinesunday

irvinesunday: I thought about it, but I was avoiding growing this function by adding an extra optional parameter (which will only be utilized by delete methods) and conditional checks here. I also wanted to be explicit that we are not using success status code range for `delete` methods.

baywet: thanks for the additional context here. Would it have been another parameter since it's an extension method on operation and thus it has access to the operation properties?

irvinesunday: But you can't infer the `OperationType` from the `operation` object properties. This is in the `Operations` property dictionary key of an `OpenApiPathItem` object: https://github.com/microsoft/OpenAPI.NET/blob/df1b405e537d0f3c2a829b6a80fef82959d0eab5/src/Microsoft.OpenApi/Models/OpenApiPathItem.cs#L29
`Operation` properties: https://github.com/microsoft/OpenAPI.NET/blob/df1b405e537d0f3c2a829b6a80fef82959d0eab5/src/Microsoft.OpenApi/Models/OpenApiOperation.cs#L119-L131

baywet: ah right, I forgot about the dictionary pattern, which is a bit painful to carry the context around. Fine, leave it as is for now, thanks for taking the time to discuss this.

irvinesunday: No worries.

Issue #365, opened 2023-03-30
Namespace stripping for functions/actions.


andrueastman:
Related to 
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1691
RunningHuntingQuery is an action and only works with the full namespace or the entire namespace stripped. not `security.RunningHuntingQuery`
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1787
getDirectRoutingCalls is an action and only works with the full namespace or the entire namespace stripped. not `callRecords.getDirectRoutingCalls`

Stripping a specific prefix from namespace from Functions and Action can lead to invalid paths. 

Following the discussion in https://github.com/microsoft/OpenAPI.NET.OData/issues/348#issuecomment-1444244198

Does it make sense to only use aliasing for TypeSegments and strip the namespace for Functions/Actions? I'm not sure if there are cases where a function/action fails if the entire namespace is stripped off?

cc @baywet @irvinesunday

baywet:
I provided [more context here](https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1787#issuecomment-1490421830) but I don't believe we should change any of the conversion logic at this point. OData semantics clearly indicate that the method should be able to be called with the fully qualified name, the aliased name and the short method name. AGS seems to be messing up the aliased scenario and we should open an ICM on them for that reason.

PR #369, opened 2023-04-06
[Fix] Aliases or strips namespace prefixes from segment names when and where applicable

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/365

This PR:
- Aliases operation or type cast segments when a namespace alias is available in the model.
- If the above is not possible, we strip off the namespace prefixes from operation segments when the value of the `NamespacePrefixToStripForInMethodPaths` matches the namespace of the operation. If these values doesn't match, we don't strip off the namespace.
- Updates unit tests.
- Updates release notes.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
@@ -378,7 +378,9 @@ namespace Microsoft.OpenApi.OData.Common
                     // Alias operation segment name 
                     segmentName = namespaceAlias.TrimEnd('.') + "." + element.Name;
                 }
-                else if (element.Namespace.Equals(settings.NamespacePrefixToStripForInMethodPaths, StringComparison.OrdinalIgnoreCase))
+                
+                if (settings.NamespacePrefixToStripForInMethodPaths != null && 
+                    element.Namespace.Equals(settings.NamespacePrefixToStripForInMethodPaths, StringComparison.OrdinalIgnoreCase))
                 {
                     // Strip specified namespace from operation segment name                  
                     segmentName = element.Name;
--- a/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------
+// ------------------------------------------------------------
 //  Copyright (c) Microsoft Corporation.  All rights reserved.
 //  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // ------------------------------------------------------------
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OData.Edm;
 using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.OData.Edm;
 using Microsoft.OpenApi.OData.Vocabulary.Capabilities;
 
 namespace Microsoft.OpenApi.OData.Common
@@ -22,7 +23,7 @@ namespace Microsoft.OpenApi.OData.Common
         {
             Utils.CheckArgumentNull(structuredType, nameof(structuredType));
             Utils.CheckArgumentNull(edmModel, nameof(edmModel));
-            if(structuredType is not IEdmSchemaElement schemaElement) throw new ArgumentException("The type is not a schema element.", nameof(structuredType));
+            if (structuredType is not IEdmSchemaElement schemaElement) throw new ArgumentException("The type is not a schema element.", nameof(structuredType));
 
             IEnumerable<IEdmSchemaElement> derivedTypes = edmModel.FindAllDerivedTypes(structuredType).OfType<IEdmSchemaElement>();
 
@@ -32,12 +33,12 @@ namespace Microsoft.OpenApi.OData.Common
             }
 
             OpenApiSchema schema = new()
-			{
+            {
                 OneOf = new List<OpenApiSchema>()
             };
 
             OpenApiSchema baseTypeSchema = new()
-			{
+            {
                 UnresolvedReference = true,
                 Reference = new OpenApiReference
                 {
@@ -50,7 +51,7 @@ namespace Microsoft.OpenApi.OData.Common
             foreach (IEdmSchemaElement derivedType in derivedTypes)
             {
                 OpenApiSchema derivedTypeSchema = new()
-				{
+                {
                     UnresolvedReference = true,
                     Reference = new OpenApiReference
                     {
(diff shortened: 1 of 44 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -341,5 +343,52 @@ internal static string GenerateODataTypeCastPathOperationIdPrefix(ODataPath path
...
+            }         
+
+            if (element is IEdmStructuredType)
+            {                
+                if (settings.EnableAliasForTypeCastSegments && !string.IsNullOrEmpty(namespaceAlias))
+                {
+                    // Alias type cast segment name
+                    segmentName = namespaceAlias.TrimEnd('.') + "." + element.Name;
+                }
+            }
```
baywet, line 375:
```suggestion
            if (element is IEdmStructuredType && settings.EnableAliasForTypeCastSegments && !string.IsNullOrEmpty(namespaceAlias))
            {
                // Alias type cast segment name
                segmentName = namespaceAlias.TrimEnd('.') + "." + element.Name;
            }
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataTypeCastSegment.cs
```diff
@@ -53,20 +50,11 @@ public override IEnumerable<IEdmVocabularyAnnotatable> GetAnnotables()
...
-
-        if(settings.EnableAliasForTypeCastSegments && !string.IsNullOrEmpty(namespaceAlias))
-        {
-            return namespaceAlias.TrimEnd('.') + "." + StructuredType.FullTypeName().Split(new char[] { '.' }, StringSplitOptions.RemoveEmptyEntries).Last();
-        }           
-
-        return StructuredType.FullTypeName();
+        return element != null && _model != null
+            ? EdmModelHelper.StripOrAliasNamespacePrefix(element, settings, _model)
+            : (element?.FullName());
```
baywet, line 58:
```suggestion
        return StructuredType is IEdmSchemaElement element && _model != null
            ? EdmModelHelper.StripOrAliasNamespacePrefix(element, settings, _model)
            : (StructuredType.FullName());
```

src/Microsoft.OpenApi.OData.Reader/Edm/ODataTypeCastSegment.cs
```diff
@@ -53,20 +50,11 @@ public override IEnumerable<IEdmVocabularyAnnotatable> GetAnnotables()
     public override string GetPathItemName(OpenApiConvertSettings settings, HashSet<string> parameters)
     {
         Utils.CheckArgumentNull(settings, nameof(settings));
-        string namespaceName = string.Empty;
-        string namespaceAlias = string.Empty;
+        
+        IEdmSchemaElement element = StructuredType as IEdmSchemaElement;
```
baywet, line 54:
```suggestion
```

baywet (changes requested):

baywet (approved):
Thanks for the quick turnaround on this one. I'd like another review from @andrueastman before we merge it.

baywet (approved):
Thanks for the quick turnaround on this one. I'd like another review from @andrueastman before we merge it.

baywet (approved):

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -341,5 +343,49 @@ internal static string GenerateODataTypeCastPathOperationIdPrefix(ODataPath path
...
+
+            if (element is IEdmStructuredType && settings.EnableAliasForTypeCastSegments && !string.IsNullOrEmpty(namespaceAlias))
+            {
+                // Alias type cast segment name
+                segmentName = namespaceAlias.TrimEnd('.') + "." + element.Name;
+            }
+            
+            if (element is IEdmOperation)
+            {                
+                if (settings.EnableAliasForOperationSegments && !string.IsNullOrEmpty(namespaceAlias))
```
andrueastman, line 376: Just to confirm here, do we want the `EnableAliasForOperationSegments` to be mutually exclusive from the `NamespacePrefixToStripForInMethodPaths`. 

Currently, as the `microsoft.graph` root namespace has the `graph` alias, enabling `EnableAliasForOperationSegments` would mean methods have the `graph` prefix irrespective of the `NamespacePrefixToStripForInMethodPaths` setting, making it not possible to have aliasing in the non-root namespaces while at the same time stripping off segments in the root namesapace.

irvinesunday: > making it not possible to have aliasing in the non-root namespaces while at the same time stripping off segments in the root namesapace.

I've not understood the correlation in this sentence, however, it is true that with the current conditions, if the root namespace has alias `graph`, then operations will be aliased and `NamespacePrefixToStripForInMethodPaths` will have no effect. I propose that they indeed need to maintain their mutual exclusivity, however, the check can be re-structured differently, for example: 
If you have a segment that has been aliased in the condition `EnableAliasForOperationSegments==true` e.g. `graph.add`, the condition `NamespacePrefixToStripForInMethodPaths != null` can check whether the _n-1_ segments are contained in the `NamespacePrefixToStripForInMethodPaths` --> `"microsoft.graph".Contains("graph")` ? Which will successfully remove the aliased prefix. And in the example, `"microsoft.graph".Contains("microsoft.graph.security")` ? which will not successfully remove the prefix.

```CSharp
if (settings.EnableAliasForOperationSegments && !string.IsNullOrEmpty(namespaceAlias))
{
    // Alias operation segment name 
    segmentName = namespaceAlias.TrimEnd('.') + "." + element.Name;
}

if (element.Namespace.Contains(settings.NamespacePrefixToStripForInMethodPaths))
{
    // Strip specified namespace from operation segment name                  
    segmentName = element.Name;
}
```

andrueastman (approved):

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -341,5 +343,51 @@ internal static string GenerateODataTypeCastPathOperationIdPrefix(ODataPath path
...
+                {
+                    // Alias operation segment name 
+                    segmentName = namespaceAlias.TrimEnd('.') + "." + element.Name;
+                }
+                
+                if (settings.NamespacePrefixToStripForInMethodPaths != null &&
```
baywet, line 382:
```suggestion
                else if (!string.IsNullOrEmpty(settings.NamespacePrefixToStripForInMethodPaths) && 
```

Didn't we decide not to strip prefix when we have an alias? without the else both conditions could apply even though their body is definitive as of today

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
andrueastman: @baywet I believe both conditions would only apply when the method is in the root namespace and the root namespace has an alias as well. 
In which case, we would rather have no prefix. Changing to the `else if` would lead to `microsoft.graph.add` to become `graph.add`(root namespace has the alias `graph`) rather than `add`. Yes?

andrueastman: +1
 
cc @irvinesunday

irvinesunday: Coolio

Issue #323, opened 2022-12-21
Missing paths on derived/odata cast paths


maisarissi:
The issue https://github.com/microsoft/OpenAPI.NET.OData/pull/302 was raised to fix missing paths like https://github.com/microsoftgraph/msgraph-metadata/issues/179 and https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/338.

However, after some validation, @peombwa confirmed that the new implementation when setting `ExpandDerivedTypesNavigationProperties` to true is generating many invalid/unsupported paths.

Is there is no better way to identify which paths should be generated, should we have a custom annotation so API owners could add it in the API description?

Related to https://github.com/microsoft/OpenAPI.NET.OData/issues/268

irvinesunday:
One way of fixing the invalid paths here will involve implementing the new annotations on actions/functions: https://issues.oasis-open.org/browse/ODATA-1538#
- `Org.OData.Capabilities.V1.ExplicitOperationBindings`
- `Org.OData.Capabilities.V1.RequiresExplicitBinding`

These will help control the generation of operations bound to the derived/OData type cast segments.
This will also help resolve: https://github.com/microsoft/OpenAPI.NET.OData/issues/232

maisarissi:
Would love to get perspectives from @darrelmiller @sebastienlevert

sebastienlevert:
Now that the spec seems applied, it feels like the right path. I don't have the context but it seems to me to be the way forward. What is the broader impact?

Issue #232, opened 2022-06-10
Bound operations not appended to non-contained navigation property paths


irvinesunday:
Related to https://github.com/microsoft/OpenAPI.NET.OData/issues/201

With this bound operation:
![image](https://user-images.githubusercontent.com/40403681/173085653-2b8c916d-1632-40c6-8480-062fc47a013d.png)

and given this non-contained navigation property:
![image](https://user-images.githubusercontent.com/40403681/173085516-e8a7bb8b-23f4-433f-96d7-d7e2c11128b0.png)

and given a schema namespace value of `microsoft.graph`

We expect to generate paths similar to below:
`.../followedSites/microsoft.graph.add`

This is not the case with the current version because we check the containment of the navigation property [here](https://github.com/microsoft/OpenAPI.NET.OData/blob/079f8ea2dfa227325a2ff7905cab86f15ee2347f/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L682-L685).
According to the spec. doc., see [ref](http://docs.oasis-open.org/odata/odata/v4.0/os/part1-protocol/odata-v4.0-os-part1-protocol.html#_Toc372793738):

```
Actions and Functions MAY be bound to an entity type, primitive type, complex type, or a collection...
The namespace- or alias-qualified name of a bound operation may be appended to any URL that identifies 
a resource whose type matches, or is derived from, the type of the binding parameter. 
```
This may be applied to any navigation property segment as long as its type matches the type of the binding parameter.

baywet:
Thanks for documenting this in details. I'm torn here.
Historically we've decided to only bind operations to contained navigation properties, entity sets or singletons. Or in other words to only have operations on the entities canonical path. This is to avoid expanding too many paths and growing the description too much. It makes sense since why would one call an operation on a "deeper path" when a shallower one is already available?

This case is an interesting example, and poor API design IMHO. The followed sites should instead be a collection of "followedSite", contained. FollowedSite is a new entity type that semantically belongs to the user (is contained), has a non-contained reference to site (existing entity), and potentially other properties like "followedDate", "followedReason", etc...

But since the API is not getting fixed any time soon, we probably out to make an opiniated decision here.
What would be the consequence of setting contains target true for that property?
Can we rely on an annotation/attribute that would signal "bind operations even though this is not contained"? 
Do we have other cases like this one?

baywet:
[...]

PR #378, opened 2023-04-28
[Feature] Adds support for `RequiresExplicitBinding` and `ExplicitOperationBindings` annotations for operations

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/323, https://github.com/microsoft/OpenAPI.NET.OData/issues/232

**This PR:**
- Uses the annotations `Org.OData.Core.V1.RequiresExplicitBinding` in operations and `Org.OData.Core.V1.ExplicitOperationBindings` in model elements to associate which operations are to be generated for which elements.
- Updates the unit and integration tests.
- Updates the release notes.

**How it works**

Take for example the operation:
```XML
<Action Name="getMemberObjects" IsBound="true">
  <Annotation Term="Org.OData.Core.V1.RequiresExplicitBinding"/>
  <Parameter Name="bindingParameter" Type="graph.directoryObject" Nullable="false" />
  <Parameter Name="securityEnabledOnly" Type="Edm.Boolean" />
  <ReturnType Type="Collection(Edm.String)" Nullable="false" Unicode="false" />
</Action>
```
The `bindingParameter` is of type `graph.directoryObject`. This means that all model elements of this type (including all of its derived types) will have the operation `getMemberObjects` appended to their paths. In order to restrict the generation of this operation only to certain instances of the binding type - `directoryObject`, we append the `<Annotation Term="Org.OData.Core.V1.RequiresExplicitBinding"/>` annotation to the operation (as shown above). 
> All the operations which bind to `directoryObject`, for example, may need to be annotated with this annotation to avoid generating irrelevant operations to all of its derived types..

The navigation property `deletedItems` is an instance of type `Collection(graph.directoryObject)`:
`<NavigationProperty Name="deletedItems" Type="Collection(graph.directoryObject)" ContainsTarget="true" />` 
Ordinarily it would have the path `/directory/deletedItems/{directoryObject-id}/microsoft.graph.getMemberGroups` which is invalid. The only valid operation for this navigation property path would be `/directory/deletedItems/{directoryObject-id}/microsoft.graph.restore`.

To restrict all the operations bound to instances of `directoryObject` from being generated as operations on the `deletedItems` navigation property and only allow the `restore` operation, we would create an `ExplicitOperationBindings` to the navigation property as such:

```XML
<NavigationProperty Name="deletedItems" Type="Collection(graph.directoryObject)" ContainsTarget="true">
  <Annotation Term="Org.OData.Core.V1.ExplicitOperationBindings">
    <Collection>
      <String>microsoft.graph.restore</String>
    </Collection>
  </Annotation>
</NavigationProperty>
```

The entity type `directoryObject` will need to invoke all the operations that are bound to it and since all the operations which bind to `directoryObject` have been annotated with the annotation `Org.OData.Core.V1.RequiresExplicitBinding`, a corresponding `Org.OData.Core.V1.ExplicitOperationBindings` annotations with the collection of allowable operations need to be specified. Specifying this annotation on the entity type itself would automatically propagate to all its derived types. Which is not what we want. Therefore, this corresponding annotation will need to be annotated on the navigation source instance of this entity type. For example:

```XML
<Annotations Target="microsoft.graph.GraphService/directoryObjects">
  <Annotation Term="Org.OData.Core.V1.ExplicitOperationBindings">
    <Collection>
      <String>microsoft.graph.checkMemberGroups</String>
      <String>microsoft.graph.checkMemberObjects</String>
      <String>microsoft.graph.getMemberGroups</String>
      <String>microsoft.graph.getMemberObjects</String>
    </Collection>
  </Annotation>
</Annotations>
```
The above will yield the paths:
```
/directoryObjects/{directoryObject-id}/microsoft.graph.checkMemberGroups
/directoryObjects/{directoryObject-id}/microsoft.graph.checkMemberObjects
/directoryObjects/{directoryObject-id}/microsoft.graph.getMemberGroups
/directoryObjects/{directoryObject-id}/microsoft.graph.getMemberObjects
```

Unless a corresponding derived type of `directoryObject` gets annotated with the `Org.OData.Core.V1.ExplicitOperationBindings`, these operations will not be created for paths of this derived type.

The net result from all this should be a significant reduction in the number of invalid operation paths and a reduction in the size of the OpenAPI document and consequently our SDKs. The appropriate annotations will need to be added to the CSDL through XSLT initially and later by workload teams to realise this.

**NB:** The default behaviour is preserved (all operations generated for the bound entity types) when none of these annotations are specified anywhere.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
@@ -798,7 +798,12 @@ namespace Microsoft.OpenApi.OData.Edm
                         
                         if (annotatable != null && !EdmModelHelper.IsOperationAllowed(_model, edmOperation, annotatable))
                         {
-                            continue;
+                            // Check whether the navigation source is allowed to have an operation on the entity type
+                            annotatable = (secondLastPathSegment as ODataNavigationSourceSegment)?.NavigationSource as IEdmVocabularyAnnotatable;
+                            if (annotatable != null && !EdmModelHelper.IsOperationAllowed(_model, edmOperation, annotatable))
+                            {
+                                continue;
+                            }                            
                         }
 
                         ODataPath newPath = subPath.Clone();
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
@@ -730,27 +730,7 @@ namespace Microsoft.OpenApi.OData.Edm
                 }
 
                 var firstEntityType = bindingType.AsEntity().EntityDefinition();
-
-                var requiresExplicitBinding = _model.FindVocabularyAnnotations(edmOperation).FirstOrDefault(x => x.Term.Name == CapabilitiesConstants.RequiresExplicitBindingName);
                 
-                if (requiresExplicitBinding != null)
-                {
-                    _explicitOperationBindings.TryGetValue(firstEntityType.Name, out var boundOperations);
-                    if (!(boundOperations?.Any() ?? false))
-                    {
-                        boundOperations = _model.GetCollection(firstEntityType, CapabilitiesConstants.ExplicitOperationBindings)?.ToList();
-                    }
-
-                    if (boundOperations == null || !boundOperations.Contains(edmOperation.FullName()))
-                    {
-                        continue;
-                    }
-                    else
-                    {
-                        _explicitOperationBindings[firstEntityType.Name] = boundOperations;
-                    }
-                }
-
                 bool filter(IEdmNavigationSource z) =>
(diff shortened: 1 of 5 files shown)


src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -34,6 +34,8 @@ public class ODataPathProvider : IODataPathProvider
         private readonly IDictionary<IEdmEntityType, IList<ODataPath>> _dollarCountPaths =
            new Dictionary<IEdmEntityType, IList<ODataPath>>();
 
+        private Dictionary<string, IList<string>> _explicitOperationBindings = new();
```
andrueastman, line 37: Remove unused varaible
```suggestion
```

andrueastman (commented):
Thanks for this @irvinesunday

andrueastman (approved):

Issue #416, opened 2023-08-31
Bitwise/flagged enums information


andrueastman:
Related to https://github.com/microsoft/kiota/issues/3237

This is issue is a placeholder to implment proposed extension in the issue

PR #417, opened 2023-09-01
Pass Bitwise Enum information

### LINKED ISSUE #258: Support for enum flags
Taking a look at the metadata for the `SourceType` enum in the graph metadata, there exists the `IsFlags` property which is set to true.

```xml
      <EnumType Name="sourceType" IsFlags="true">
        <Member Name="mailbox" Value="1" />
        <Member Name="site" Value="2" />
      </EnumType>
```
The `flags` attribute is not propagated to the converted openApi description in any way at the moment and also information on the values is also lost in the conversion process. 

```yaml
    microsoft.graph.security.sourceType:
      title: sourceType
      enum:
        - mailbox
        - site
        - unknownFutureValue
      type: string
```

This poses a challenge to Kiota at the moment as there is no signal for the generator to generate an appropriate Flag property. 

**Notes**
- According to https://github.com/OAI/OpenAPI-Specification/issues/348#issuecomment-336194030 enums could be better described to cater for values by conforming to the json schema
- Some extensions like `x-enumFlags` to denote flagged enums
- AutoRest uses `[x-ms-enum](https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-enum)` to denote values and fields.

Related to https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/496

cc @baywet @darrelmiller

andrueastman:
Closes https://github.com/microsoft/OpenAPI.NET.OData/issues/416 Closes #258

Related to 
Adds the `x-ms-enum-flags` extension to description to enable generation of bitwise enums(https://github.com/microsoft/OpenAPI/pull/8). 

Also fixes XUnit warnings in tests related to using `Assert.Single` for checking collections with only one item.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
@@ -262,6 +262,11 @@ namespace Microsoft.OpenApi.OData
         public bool AddEnumDescriptionExtension { get; set; } = false;
 
         /// <summary>
+        /// Gets/sets a value indicating whether or not to add a "x-ms-enum-flags" extension to the enum type schema.
+        /// </summary>
+        public bool AddEnumFlagsExtension { get; set; } = true;
+        
+        /// <summary>
         /// Gets/sets a value indicating whether the error responses should be described as a default response or as 4XX and 5XX error responses.
         /// </summary>
         public bool ErrorResponsesAsDefault { get; set; } = true;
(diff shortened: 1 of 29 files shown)


src/Microsoft.OpenApi.OData.Reader/OpenApiConvertSettings.cs
```diff
@@ -261,6 +261,11 @@ public string PathPrefix
         /// </summary>
         public bool AddEnumDescriptionExtension { get; set; } = false;
 
+        /// <summary>
+        /// Gets/sets a value indicating whether or not to add a "x-ms-enum-flags" extension to the enum type schema.
+        /// </summary>
+        public bool AddEnumFlagsExtension { get; set; } = true;
```
baywet, line 267: you also need to update the clone method

baywet (changes requested):

baywet (approved):
Thanks for making the change!

Issue #429, opened 2023-10-03
Bound operations not appended to non-contained navigation property paths with ExplicitBinding annotations


andrueastman:
Follow up to https://github.com/microsoft/OpenAPI.NET.OData/issues/232

Related to https://github.com/microsoftgraph/msgraph-metadata/issues/449

Setting `ContainsTarget="false"` on a navigation property prevents the enabling of a function/action on nav property due to the check below preventing the generation of the path because it doesn't allow for the evaluation of the ExplicitBinding annotation.

https://github.com/microsoft/OpenAPI.NET.OData/blob/d9474cff6ae35c9a4a7ccb330ba517b64f567d2b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs#L828

Given the original intention of https://github.com/microsoft/OpenAPI.NET.OData/issues/232, the generation of a path when contains `ContainsTarget="false"` should take into consideration that nav property has the explicit operation binding. 

In summary, we should have the following scenarios, 

| Scenario| ContainsTarget | ExplicitBinding present | Generate? |
|--------|--------|--------|--------|
| 1 | False | False | **NO** |
| 2 | False | True | **YES** (missing scenario) |
| 3 | True | False | **YES** (already works, as the containsTarget allows this )|
| 4 | True | True | **YES** (already works, as the containsTarget allows this)|

This blocks adding paths to nav properties where contains target is false as seen in https://github.com/microsoftgraph/msgraph-metadata/issues/449

PR #430, opened 2023-10-03
Use containment together with `RequiresExplicitBinding` annotation to check whether to append bound operations to navigation properties

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/429

This PR:
- Checks whether a bound operation has the annotation `Org.OData.Core.V1.RequiresExplicitBinding` before falling back on the containment check for the target navigation property.
  - If the above annotation is present on the operation, then the `Org.OData.Core.V1.ExplicitOperationBindings` will need to be provided on the target navigation property with the said operation contained in the annotation collection for this method to be allowed. 
  - If the above annotation is absent, then the containment of the navigation property will determine whether or not the method will be allowed. If the containment is true, then the method will be allowed, else, it won't.


baywet (dismissed):

src/OoasGui/MainForm.cs
```diff
@@ -52,6 +52,17 @@ public MainForm()
 
             csdlRichTextBox.WordWrap = false;
             oasRichTextBox.WordWrap = false;
+
+
+            verifyEdmModelcheckBox.Checked = false;
```
andrueastman, line 57: Probably need to double check if this change is intended as well.

Issue #442, opened 2023-11-03
[Fix] Update `operationId`s of navigation property paths with OData type cast segments


irvinesunday:
Recently, we enable the expansion of derived types navigation properties [here](https://github.com/microsoftgraph/msgraph-metadata/pull/476).
This created paths like below:

```YAML
'/deviceAppManagement/mobileApps/{mobileApp-id}/microsoft.graph.androidStoreApp/categories/{mobileAppCategory-id}':
    description: Provides operations to manage the categories property of the microsoft.graph.mobileApp entity.
    get:
      tags:
        - deviceAppManagement.mobileApp
      summary: Get categories from deviceAppManagement
      description: The list of categories for this app.
      operationId: deviceAppManagement.mobileApps.GetCategories
```
Since we only generate operationIds of navigation property paths based on the navigation property segments, we end up with duplicate operationIds for 'similar' navigation property paths which only differ on the OData type cast segment, e.g.:

```YAML
'/deviceAppManagement/mobileApps/{mobileApp-id}/microsoft.graph.iosLobApp/categories/{mobileAppCategory-id}':
    description: Provides operations to manage the categories property of the microsoft.graph.mobileApp entity.
    get:
      tags:
        - deviceAppManagement.mobileApp
      summary: Get categories from deviceAppManagement
      description: The list of categories for this app.
      operationId: deviceAppManagement.mobileApps.GetCategories
```

We need to update the `operationId`s so as to include the OData type cast segment name to differentiate the above two scenarios.

PR #443, opened 2023-11-03
[Fix] Updates `operationId`s of navigation property paths with OData type cast segments

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/442

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
@@ -168,6 +168,15 @@ namespace Microsoft.OpenApi.OData.Common
             {
                 if (segment == navPropSegments.Last())
                 {
+                    // If there exists an OData type cast segment present in the path,
+                    // this segement identifier needs to be added to the operation id
+                    ODataTypeCastSegment typeCastSegment = path.Segments.OfType<ODataTypeCastSegment>().LastOrDefault();
+                    if (typeCastSegment != null && path.Kind == ODataPathKind.NavigationProperty)
+                    {
+                        IEdmSchemaElement schemaElement = typeCastSegment.StructuredType as IEdmSchemaElement;
+                        items.Add("As" + Utils.UpperFirstChar(schemaElement.Name));
+                    }
+                    
                     items.Add(segment.NavigationProperty.Name);
                     break;
                 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -161,19 +161,25 @@ internal static IList<string> RetrieveNavigationPropertyPathsOperationIdSegments
...
-                    items.Add(segment.NavigationProperty.Name);
-                    break;
+                    items.Add(navPropSegment.NavigationProperty.Name);
                 }
-                else
+                else if (segment is ODataTypeCastSegment typeCastSegment && path.Kind == ODataPathKind.NavigationProperty)
                 {
-                    items.Add(segment.NavigationProperty.Name);
+                    // Only the last OData type cast segment identifier is added to the operation id
+                    items.Remove(previousTypeCastSegmentId);
```
andrueastman, line 179: Should we check if `previousTypeCastSegmentId` is null before trying to remove. There looks to be a possible path if its the first time entering the block.

irvinesunday: There's no side-effect of `IEnumerable.Remove(null)`. No item in the collection will be removed, if there are any.

src/Microsoft.OpenApi.OData.Reader/Common/EdmModelHelper.cs
```diff
@@ -161,19 +161,25 @@ internal static IList<string> RetrieveNavigationPropertyPathsOperationIdSegments
                 navigationSource.Name
             };
 
-            IEnumerable<ODataNavigationPropertySegment> navPropSegments = path.Segments.Skip(1).OfType<ODataNavigationPropertySegment>();
-            Utils.CheckArgumentNull(navPropSegments, nameof(navPropSegments));
+            // For navigation property paths with odata type cast segments
+            // the OData type cast segments identifiers will be used in the operation id
+            IEnumerable<ODataSegment> segments = path.Segments.Skip(1).Where(s => s is ODataNavigationPropertySegment || s is ODataTypeCastSegment);
```
baywet, line 166:
```suggestion
            IEnumerable<ODataSegment> segments = path.Segments.Skip(1).Where(static s => s is ODataNavigationPropertySegment || s is ODataTypeCastSegment);
```

baywet (changes requested):

baywet (approved):

timayabi2020 (approved):

andrueastman (approved):

Issue #197, opened 2022-03-11
$expand is reported as enum while it is actually string


sherlock1982:
Current behavior:
$expand is enum.

Expected behavior:
$expand is string because you can have complex expressions in $expand like for example `$expand=Groups($expand=Members)`

darrelmiller:
If we updated the schema to be like this it would allow us to provide a list of properties that are expandable but also indicate that each expand property could be a more complex syntax.

```yaml
          schema:
            uniqueItems: true
            type: array
            items:
              anyOf:
                - enum:
                    - '*'
                    - serviceAnnouncement
                - type: string
              type: string
```

Thoughts?

baywet:
We do have a hard dependency on that though, GE relies on this to auto completion.
We need to tie this new behaviour to a conversion setting switch (true by default, false for GE) "EmitExpandSelectAndOrderByAsStrings" (or a better name)

andrueastman:
> I propose we simply project those three as arrays of strings instead. Which will have the additional benefit of making Microsoft Graph descriptions quite smaller too.

+1

PR #499, opened 2024-03-01
[Fix] Uses convert setting to toggle between generating query options schemas of type string array or enums

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/197

This PR:
- Proposes a new convert setting: `UseStringArrayForQueryOptionsSchema` for toggling between generating schemas 
of query options of type string array or enums.
- Uses the above convert setting when generating schemas of query options: `$select`, `$expand` and `$orderby`
- Updates release notes


baywet (approved):
Thanks for the quick turnaround on this one!
Once this is released, can you please release hidi as well?
(and update GE convert settings in the metadata repo to set this new switch to false)

irvinesunday:
> Thanks for the quick turnaround on this one! Once this is released, can you please release hidi as well? (and update GE convert settings in the metadata repo to set this new switch to false)

Will sure do 👍🏽

Issue #405, opened 2023-07-20
Add the ability to change the request or response body content type based on an annotation


baywet:
[OneNote](https://learn.microsoft.com/en-us/graph/onenote-create-page#example-request) leverages multipart/form to create/update one note pages so the page, attachments and more can be pushed in a single call.

Currently the resulting OpenAPI description contains a basic JSON input/output for that endpoint which is incorrect.

We should:
1. identify a [capability annotation](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md) that could carry the information in the metadata
2. if none exist, define a new one
3. implement parsing it
4. implement the change in projection based on the metadata
5. add transforms in the metadata repo

baywet:
Here is what the result should look like (for the request body) in the case of OneNote

```yaml
openapi: 3.0.1
info:
  title: Example
  description: Example
  version: 1.0.1
servers:
  - url: https://example.org
paths:
  /directoryObject:
    post:
      requestBody:
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                id:
                  type: string
                  format: uuid
                address:
                  $ref: '#/components/schemas/address'
                profileImage:
                  type: string
                  format: binary
            encoding:
              id:
                contentType: text/plain
              address:
                contentType: application/json
              profileImage:
                contentType: image/png
        responses:
          '204':
            content:
              application/json:
                schema:
                  type: string
components:
  schemas:
    address:
      type: object
      properties:
        street:
          type: string
        city:
          type: string
```

baywet:
Thanks for the additional information.
How do you differentiate between the request and the response media type using that annotation? What if the media type is specific to an operation?

darrelmiller:
I was going to suggest:
- SupportedRequestTypes
- AcceptableResponseTypes

but I think I like @baywet's suggestions better.

Have @garethj-msft and @MaximRouiller seen this proposal?

MaximRouiller:
We haven't. Reading this now. Give me some time to ingest the information.

MaximRouiller:
Alright, I have a few questions. 

* Are we trying to specify the exact MIME of the `Edm.Stream`? 
* I don't think setting the acceptable MIME type on the collection of pages make sense since the content of individual pages/resources of a oneNote payload could technically be different MIME types. I would see that definition belonging closer to the `<Property Name="content" Type="Edm.Stream" />` property. 

[...]

PR #500, opened 2024-03-02
[Feature] Adds ability to change the response or request body content media type based on custom annotation properties

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/405

This PR:
- Introduces two new custom annotation properties: `RequestContentTypes` and `ResponseContentTypes` to help provide means of generating expected request body and response content media types for a given endpoint. The current default is `application/json`
- Updates generation of request body and response content media types of entity sets whose entities have media streams to either generate the annotated media type from the annotated `Org.OData.Core.V1.AcceptableMediaTypes` or default to `application/octet-stream` and not generate the `application/json` media type.
- Adds and updates tests to validate the above.
- Updates release notes.


andrueastman (approved):
👍🏼

Issue #507, opened 2024-03-19
Duplicate operation Ids ``identity.authenticationEventsFlows.AsOnAttributeCollectionExternalUsersSelfServiceSignUp.DeleteRefAttributes``


timayabi2020:
1. ``/identity/authenticationEventsFlows/{authenticationEventsFlow-id}/externalUsersSelfServiceSignUpEventsFlow/onAttributeCollection/onAttributeCollectionExternalUsersSelfServiceSignUp/attributes/$ref``.
2. ``/identity/authenticationEventsFlows/{authenticationEventsFlow-id}/externalUsersSelfServiceSignUpEventsFlow/onAttributeCollection/onAttributeCollectionExternalUsersSelfServiceSignUp/attributes/{identityUserFlowAttribute-id}/$ref``

The above paths have duplicate operation ids causing PowerShell snippet generation failure as reported [here](https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/1996).

<img width="668" alt="image" src="https://github.com/microsoft/OpenAPI.NET.OData/assets/10947120/e856d140-e1e0-41b9-8fb5-ce21e064f8e6">

PR #513, opened 2024-03-25
Generates unique DELETE operation ids of `$ref` paths for indexed collection navigation properties

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/507

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyOperationHandler.cs
@@ -42,6 +42,11 @@ namespace Microsoft.OpenApi.OData.Operation
         protected bool LastSegmentIsKeySegment { get; private set; }
 
         /// <summary>
+        /// Gets a bool value indicating whether the second last segment in a $ref path is a key segment
+        /// </summary>
+        protected bool SecondLastSegmentIsKeySegment { get; private set; }
+
+        /// <summary>
         /// Gets a bool value indicating whether the last segment is a $ref segment.
         /// </summary>
         protected bool LastSegmentIsRefSegment { get; private set; }
@@ -56,6 +61,7 @@ namespace Microsoft.OpenApi.OData.Operation
 
             LastSegmentIsKeySegment = path.LastSegment is ODataKeySegment;
             LastSegmentIsRefSegment = path.LastSegment is ODataRefSegment;
+            SecondLastSegmentIsKeySegment = Path.Segments.Reverse().Skip(1).Take(1).Single().Kind == ODataSegmentKind.Key;
             NavigationProperty = path.OfType<ODataNavigationPropertySegment>().Last().NavigationProperty;
 
             IEdmEntitySet entitySet = NavigationSource as IEdmEntitySet;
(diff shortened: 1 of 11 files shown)


src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyOperationHandler.cs
```diff
@@ -55,7 +60,8 @@ protected override void Initialize(ODataContext context, ODataPath path)
             NavigationSource = navigationSourceSegment.NavigationSource;
 
             LastSegmentIsKeySegment = path.LastSegment is ODataKeySegment;
-            LastSegmentIsRefSegment = path.LastSegment is ODataRefSegment;
+            LastSegmentIsRefSegment = path.LastSegment is ODataRefSegment;
+            SecondLastSegmentIsKeySegment = Path.Segments.Reverse().Skip(1).Take(1).Single().Kind == ODataSegmentKind.Key;
```
baywet, line 64: should we check that we have at least 2 elements in the segments collection before calling skip/take/single?

src/Microsoft.OpenApi.OData.Reader/Operation/RefDeleteOperationHandler.cs
```diff
@@ -40,8 +40,22 @@ protected override void SetBasicInfo(OpenApiOperation operation)
             // OperationId
             if (Context.Settings.EnableOperationId)
             {
-                string prefix = "DeleteRef";
-                operation.OperationId = GetOperationId(prefix);
+                string prefix = "DeleteRef";
+                var segments = GetOperationId().Split('.').ToList();
```
baywet, line 44:
```suggestion
                var segments = GetOperationId().Split('.').ToList();
                if (segments.Length <= 0) return;
```

baywet (changes requested):

src/Microsoft.OpenApi.OData.Reader/Operation/RefDeleteOperationHandler.cs
irvinesunday: In this case... `if (segments.Count <= 0) return;`

src/Microsoft.OpenApi.OData.Reader/Operation/NavigationPropertyOperationHandler.cs
irvinesunday: We can, but I don't think we'd ever have such a scenario where we have less than 2 segments, since `/$ref` segments will always be appended to at least: 
`~navSource/col-nav-prop/{id}/$ref` or `~navSource/nav-prop/$ref`

src/Microsoft.OpenApi.OData.Reader/Operation/RefDeleteOperationHandler.cs
baywet: I think part of your comment is missing?

andrueastman: I believe the `if (segments.Length <= 0) return;` is technically unreachable code. As `Split` will still return the original string as the only item in the collection if it does not contain the split character.

baywet: right, it only works if remove empty entries is passed
https://dotnetfiddle.net/JGpHaR
Resolving

baywet (approved):

PR #523, opened 2024-04-25
Add `delete` operations for media entities

### LINKED ISSUE #522: Add `Deletability` support for media entities
Ref #https://github.com/microsoftgraph/msgraph-metadata/issues/593

millicentachieng:
Fixes #522 
This PR:
- Adds DELETE operations for media content paths unless there are annotations specifying that an entity isn't deletable
- Adds tests to check that the operation details are correct

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/MediaEntityDeleteOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/MediaEntityDeleteOperationHandler.cs
@@ -4,10 +4,7 @@ using Microsoft.OpenApi.OData.Common;
 using Microsoft.OpenApi.OData.Edm;
 using Microsoft.OpenApi.OData.Generator;
 using Microsoft.OpenApi.OData.Vocabulary.Capabilities;
-using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace Microsoft.OpenApi.OData.Operation
 {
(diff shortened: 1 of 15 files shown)


src/Microsoft.OpenApi.OData.Reader/PathItem/MediaEntityPathItemHandler.cs
```diff
@@ -50,6 +50,15 @@ protected override void SetOperations(OpenApiPathItem item)
             {
                 AddOperation(item, OperationType.Put);
             }
+
+            DeleteRestrictionsType delete = EntitySet != null
+                ? Context.Model.GetRecord<DeleteRestrictionsType>(EntitySet)
+                : Context.Model.GetRecord<DeleteRestrictionsType>(Singleton);
+
+            if (delete == null || delete.IsDeletable)
```
andrueastman, line 58: Given path is typically from a `/$value` segment, I would suggest only adding the operation if the annotation exists. This is because we may end up adding a number of paths(and increasing metadata size) that are not valid (for examples in MS graph, the deleting an message or the deleting of an attachment by the `/$value` path) which isn't useful.
This is also given that we only know of very few paths that support `delete` in the /$value segment.

src/Microsoft.OpenApi.OData.Reader/Operation/MediaEntityDeleteOperationHandler.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            base.SetBasicInfo(operation);
+        }
+
+        /// <inheritdoc/>
+        protected override void SetParameters(OpenApiOperation operation)
+        {
+            base.SetParameters(operation);
+
+            operation.Parameters.Add(new OpenApiParameter
```
andrueastman, line 75: out of curiosity, is there a reason the `"If-Match"` header is added by default?

src/Microsoft.OpenApi.OData.Reader/PathItem/MediaEntityPathItemHandler.cs
garethj-msft: Per the OData spec, DELETE should be supported on both <path_to_entity> AND <path_to_entity>/$value [here](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_DeleteaMediaEntity).

There aren't enough media entities in the Graph metadata for the size increase to be meaningful.

andrueastman: Thanks for the info!

Issue #511, opened 2024-03-22
Add support of `Edm.Untyped`


andrueastman:
Odata 4.01 supports the use of the [`Edm.Untyped`](https://docs.oasis-open.org/odata/odata/v4.01/csprd01/part3-csdl/odata-v4.01-csprd01-part3-csdl.html#_Toc469935967) which is used for situations where the OData type system is too restrictive to model the payload. 

At the moment the conversion library does not support this and will throw a `NotSupported` error and therefore validation/conversion of CSDL with this type will error out.
-https://github.com/microsoft/OpenAPI.NET.OData/blob/d48c3966595097c66f41a52cebf28b4d1c34c50e/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiSchemaGenerator.cs#L702
- https://github.com/microsoft/OpenAPI.NET.OData/blob/d48c3966595097c66f41a52cebf28b4d1c34c50e/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiExampleGenerator.cs#L182

To unlock the scenarios where an api accepts untyped data and generate a valid open description, we should generate the following descriptions.

1. Collection of untyped data i.e. Collection(Edm.Untyped) should OpenApiObject() in the **items** field.
```yaml
schema:
  type: "array"
  items: {}
```

2. Untyped data i.e.(Edm.Untyped) should be an empty `OpenApiObject`
```yaml
schema: {}
```

This will unblock scenarios such as https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1783 once https://github.com/microsoftgraph/msgraph-metadata/issues/596 is closed.

MartinM85:
Is there any progress on this? Looks easy to implement, but I'm not sure about tests.

millicentachieng:
@MartinM85 I have started working on this.

PR #531, opened 2024-05-16
Add support for Edm.Untyped

millicentachieng:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/511


test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
```diff
@@ -30590,6 +30590,11 @@
             "type": "string"
           }
         },
+        "TripData": {},
```
xuzhg, line 30593: Leave it an empty object? or use a 'type': 'object'?

millicentachieng:
```suggestion
        "TripData": { 
              "type": "object"
        },
```
@andrueastman Your thoughts?

andrueastman: We should leave it as empty. The reason being that the data could be a primitive as well(integer, string ...) or even a collection. As the type is only known at runtime so I think it would be incorrect to allude that it's an object.

andrueastman (approved):

test/Microsoft.OpenAPI.OData.Reader.Tests/Resources/TripService.OpenApi.V2.json
millicentachieng: Or we could have something like this (the list of types is not exhaustive): 
```
components:
  schemas:
    ExampleResource:
      type: object
      properties:
        id:
          type: string
        customData:
          anyOf:
            - type: string
            - type: number
            - type: integer
            - type: boolean
            - type: object
            - type: array
              items:
                anyOf:
                  - type: string
                  - type: number
                  - type: integer
                  - type: boolean
                  - type: object
```

garethj-msft: Its not practical to model out everything it could be, as it could be a collection of structs each with a heterogeneous collection that's a mixture of bools, ints and more structs. So, as @andrueastman says, it's completely unknown at design time.

My JSON schema skill isn't strong, but whatever expresses that complete lack of knowledge whether it is primitive, object or array, that is what we should do.

millicentachieng: Thanks, @garethj-msft

Issue #546, opened 2024-06-26
Some OData type cast paths are returning empty response objects and failing Hidi validation


irvinesunday:
```YAML
'/identity/authenticationEventsFlows/{authenticationEventsFlow-id}/microsoft.graph.externalUsersSelfServiceSignUpEventsFlow/onAuthenticationMethodLoadStart/microsoft.graph.onAuthenticationMethodLoadStartExternalUsersSelfServiceSignUp':
    description: Casts the previous resource to onAuthenticationMethodLoadStartExternalUsersSelfServiceSignUp.
    get:
      tags:
        - identity.authenticationEventsFlow
      summary: Get the items of type microsoft.graph.onAuthenticationMethodLoadStartExternalUsersSelfServiceSignUp in the microsoft.graph.onAuthenticationMethodLoadStartHandler collection
      operationId: identity.authenticationEventsFlows.GetOnAuthenticationMethodLoadStart.AsOnAuthenticationMethodLoadStartExternalUsersSelfServiceSignUp
      parameters:
        - name: authenticationEventsFlow-id
          in: path
          description: The unique identifier of authenticationEventsFlow
          required: true
          schema:
            type: string
          x-ms-docs-key-type: authenticationEventsFlow
      responses:
        '200': { }
        default:
          $ref: '#/components/responses/error'
```

The empty `responses` object:  `'200': { }` is causing validation error in Hidi: 

```
 Detected error during parsing: The field 'description' in 'response' object is REQUIRED. [#/paths/~1identity~1authenticationEventsFlows~1{authenticationEventsFlow-id}~1microsoft.graph.externalUsersSelfServiceSignUpEventsFlow~1onAuthenticationMethodLoadStart~1microsoft.graph.onAuthenticationMethodLoadStartExternalUsersSelfServiceSignUp/get/responses/200/description]
```

PR #547, opened 2024-06-26
Fixes empty response objects for OData type cast paths

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/546

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/ODataTypeCastGetOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/ODataTypeCastGetOperationHandler.cs
@@ -197,9 +197,33 @@ internal class ODataTypeCastGetOperationHandler : OperationHandler
 	protected override void SetResponses(OpenApiOperation operation)
     {
         if (IsSingleElement)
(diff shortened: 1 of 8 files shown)


src/Microsoft.OpenApi.OData.Reader/Operation/ODataTypeCastGetOperationHandler.cs
```diff
@@ -187,79 +195,17 @@ protected override void SetBasicInfo(OpenApiOperation operation)
...
-	}
-
-	private void SetSingleResponse(OpenApiOperation operation)
-	{
-		OpenApiSchema schema = null;
-
-		if (Context.Settings.EnableDerivedTypesReferencesForResponses)
-		{
-			schema = EdmModelHelper.GetDerivedTypesReferenceSchema(targetStructuredType, Context.Model);
-		}
```
andrueastman, line 227: Out of curiosity, is this functionality lost with the latest change? or is this not needed?

irvinesunday: Thanks for pointing this out. I had actually missed this

andrueastman (approved):

Issue #551, opened 2024-07-17
Add support for configuring composable functions generations


irvinesunday:
In order to support all the expected composable functions, we need to revert this [XSL transform](https://github.com/microsoftgraph/msgraph-metadata/pull/561) and update the conversion library to be able to toggle the depth of expanding composable functions. The initial default depth will be set to 1.

This issue is related to: https://github.com/microsoftgraph/msgraph-metadata/issues/644

PR #552, opened 2024-07-17
Adds config setting for composable functions generation

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/551

Adds the config setting: `ComposableFunctionsExpansionDepth` which provides a control for how many levels composable functions are to be expanded. The default is 1. This means composable functions will be expanded to generate an operation segment or a navigation property segment in the path, but not both. If the value is set to 2, then the latter will be allowed.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
@@ -888,25 +888,31 @@ namespace Microsoft.OpenApi.OData.Edm
                     continue;
                 }
 
-                foreach (var navProperty in returnBindingEntityType.NavigationProperties())
-                {
-                    if (convertSettings.ComposableFunctionsExpansionDepth > 0)
-                    { // Generate nav. props of composable functions
-                        ODataSegment secondLastSeg = functionPath.ElementAt(functionPath.Count - 2);
-
-                        if (convertSettings.ComposableFunctionsExpansionDepth < 2 &&
+                ODataSegment secondLastSeg = functionPath.ElementAt(functionPath.Count - 2);
+                if (convertSettings.ComposableFunctionsExpansionDepth < 2 &&
                             functionPath.LastSegment is ODataOperationSegment &&
                             secondLastSeg is ODataOperationSegment)
-                        {
-                            // Only one level of composable functions expansion allowed
-                            continue;
-                        }
-                        else
-                        {
-                            ODataPath newNavigationPath = functionPath.Clone();
-                            newNavigationPath.Push(new ODataNavigationPropertySegment(navProperty));
-                            AppendPath(newNavigationPath);
-                        }
+                {
+                    // Only one level of composable functions expansion allowed
+                    continue;
+                }
+
+                foreach (var navProperty in returnBindingEntityType.NavigationProperties())
+                {
+                    /* Get number of segments already appended after the first composable function segment
+                     */
+                    int composableFuncSegIndex = functionPath.Segments.IndexOf(
+                        functionPath.Segments.FirstOrDefault(
+                            x => x is ODataOperationSegment operationSegment &&
+                            operationSegment.Operation is IEdmFunction edmFunction &&
+                            edmFunction.IsComposable));                   
+                    int currentDepth = functionPath.Count - composableFuncSegIndex - 1;
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -890,9 +890,24 @@ private void RetrieveBoundOperationPaths(OpenApiConvertSettings convertSettings)
 
                 foreach (var navProperty in returnBindingEntityType.NavigationProperties())
                 {
-                    ODataPath newNavigationPath = functionPath.Clone();
-                    newNavigationPath.Push(new ODataNavigationPropertySegment(navProperty));
-                    AppendPath(newNavigationPath);
+                    if (convertSettings.ComposableFunctionsExpansionDepth > 0)
```
andrueastman, line 893: if statement can be moved outside the for each to avoid the repeated calculation of the branch.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -890,9 +890,24 @@ private void RetrieveBoundOperationPaths(OpenApiConvertSettings convertSettings)
...
+                    { // Generate nav. props of composable functions
+                        ODataSegment secondLastSeg = functionPath.ElementAt(functionPath.Count - 2);
+
+                        if (convertSettings.ComposableFunctionsExpansionDepth < 2 &&
+                            functionPath.LastSegment is ODataOperationSegment &&
+                            secondLastSeg is ODataOperationSegment)
+                        {
+                            // Only one level of composable functions expansion allowed
+                            continue;
```
andrueastman, line 902: Does this mean that we cannot introduce granularity like limiting the path to 3 or 4 expansions?

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: Okay, yes, thank you.

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
irvinesunday: Good feedback. I have made the generation more granular, depending on the setting

src/Microsoft.OpenApi.OData.Reader/Edm/ODataPathProvider.cs
```diff
@@ -888,11 +888,32 @@ private void RetrieveBoundOperationPaths(OpenApiConvertSettings convertSettings)
...
+                    /* Get number of segments already appended after the first composable function segment
+                     */
+                    int composableFuncSegIndex = functionPath.Segments.IndexOf(
+                        functionPath.Segments.FirstOrDefault(
+                            x => x is ODataOperationSegment operationSegment &&
+                            operationSegment.Operation is IEdmFunction edmFunction &&
+                            edmFunction.IsComposable));                   
+                    int currentDepth = functionPath.Count - composableFuncSegIndex - 1;
+
+                    if (currentDepth < convertSettings.ComposableFunctionsExpansionDepth)
```
andrueastman, line 911:
```suggestion
                    if (currentDepth < convertSettings.ComposableFunctionsExpansionDepth && composableFuncSegIndex >= 0)
```
We probably want to be sure that the search using `IndexOf` was successful in line 904

irvinesunday: Since at that point we are sure we have at least 1 operation segment that `IsComposable = true`, `composableFuncSegIndex` will never be a negative integer. However, to fool proof this, we can consider putting the safeguard when initializing the `currentDepth` variable:
`int currentDepth = (composableFuncSegIndex > 0) ?  functionPath.Count - composableFuncSegIndex - 1 : 0;`

andrueastman: Thanks for clarifying this. After taking further look, looks like we have the guarantee that the last segment will at least meet this condition. Resolving this for now ....

andrueastman (approved):

PR #561, opened 2024-08-14
Upgrade reader to .NET 8

### LINKED ISSUE #479: Modernization of source code: Update .NET package to version 8
- Take advantage of the performance improvements in .NET 8 ie runtime optimization which includes reduction in resources consumption, reduced startup times, full utilization of multicore hardware architecture to improve parallelization. 
- Debugging experience through the improved profiling tools
- JSON serialization/deserialization improvements

millicentachieng:
Closes #479

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
+++ b/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
@@ -12,7 +12,7 @@
     <Company>Microsoft</Company>
     <Title>Microsoft Open API OData Reader</Title>
     <RootNamespace>Microsoft.OpenApi.OData</RootNamespace>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net8</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


irvinesunday (approved):

src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
```diff
@@ -12,7 +12,7 @@
     <Company>Microsoft</Company>
     <Title>Microsoft Open API OData Reader</Title>
     <RootNamespace>Microsoft.OpenApi.OData</RootNamespace>
```
andrueastman, line 14: Just checking here, does this mean we want to stop supporting runtimes like net framework? 
Anyone pulling the package from a net framework project or version earlier than net8.0 won't be able to use this and could arguably be a breaking change....

irvinesunday: We may need to bump the major version here.

millicentachieng: We have bumped the major version.

Issue #564, opened 2024-08-20
Add support for retrieving multiple enum values from the `UpdateMethod` property of the `UpdateRestrictions` annotation


irvinesunday:
We need to add support for retrieving multiple enum values from the `UpdateMethod` property value of `UpdateRestrictions` annotations. 

example:

```XML
<Annotations Target="microsoft.graph.customClaimsPolicy/claims">
  <Annotation Term="Org.OData.Capabilities.V1.UpdateRestrictions">
    <Record>
      <PropertyValue Property="UpdateMethod">
        <Collection>
          <EnumMember>Org.OData.Capabilities.V1.HttpMethod/PUT Org.OData.Capabilities.V1.HttpMethod/PATCH</EnumMember>
        </Collection>              
      </PropertyValue>
    </Record>
  </Annotation>
</Annotations>
```

baywet:
Dumping some of the conversation elements:

We'll need to update these blocks (multiple places) not to be if/else, but if first condition, if second condition.
https://github.com/microsoft/OpenAPI.NET.OData/blob/46c149087f3fdcca62cca5a0590a26a4c673e2a9/src/Microsoft.OpenApi.OData.Reader/PathItem/EntityPathItemHandler.cs#L37

The enum above needs to be flaggable https://github.com/microsoft/OpenAPI.NET.OData/blob/46c149087f3fdcca62cca5a0590a26a4c673e2a9/src/Microsoft.OpenApi.OData.Reader/Vocabulary/Capabilities/UpdateRestrictionsType.cs#L161
so it matches its vocab definition
https://github.com/oasis-tcs/odata-vocabularies/blob/a7850482cc83488d37c66c4e907f6a94406e1d05/vocabularies/Org.OData.Capabilities.V1.xml#L764C23-L764C33

And we might have to update the GetEnumMember helper so it supports flags enums.

https://github.com/microsoft/OpenAPI.NET.OData/blob/46c149087f3fdcca62cca5a0590a26a4c673e2a9/src/Microsoft.OpenApi.OData.Reader/Edm/RecordExpressionExtensions.cs#L96

baywet:
Another aspect to consider when emitting the multiple operations is the operation id:
- they should not collide, including with POST when present
- PUT should be Update by default, New when POST is absent AND PATCH is present.
- PATCH should be Update
- POST should be New

This leaves a question of what should the Id be for PUT when all of POST/PATCH/PUT are present?

PR #567, opened 2024-08-20
Adds support for retrieving collection of enum values from `UpdateMethod` property of `UpdateRestrictions` annotation

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/564

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Vocabulary/Capabilities/UpdateRestrictionsType.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Vocabulary/Capabilities/UpdateRestrictionsType.cs
@@ -49,10 +49,10 @@ namespace Microsoft.OpenApi.OData.Vocabulary.Capabilities
         public bool? DeltaUpdateSupported { get; private set; }
 
         /// <summary>
-        /// Gets the value indicating the HTTP Method (PUT or PATCH) for updating an entity. 
+        /// Gets the values indicating the HTTP Method (PUT or/and PATCH) for updating an entity. 
         /// If null, PATCH should be supported and PUT MAY be supported.
         /// </summary>
-        public HttpMethod? UpdateMethod { get; private set; }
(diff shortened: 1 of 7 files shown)


src/Microsoft.OpenApi.OData.Reader/Vocabulary/Capabilities/UpdateRestrictionsType.cs
```diff
@@ -49,10 +49,10 @@ internal class UpdateRestrictionsType : IRecord
         public bool? DeltaUpdateSupported { get; private set; }
 
         /// <summary>
-        /// Gets the value indicating the HTTP Method (PUT or PATCH) for updating an entity. 
+        /// Gets the values indicating the HTTP Method (PUT and/or PATCH) for updating an entity. 
         /// If null, PATCH should be supported and PUT MAY be supported.
         /// </summary>
-        public HttpMethod? UpdateMethod { get; private set; }
+        public IList<HttpMethod> UpdateMethods { get; private set; }
```
andrueastman, line 55: I believe the ask here is to make the enum flaggable with the [Flags] attribute. https://github.com/microsoft/OpenAPI.NET.OData/blob/e6efed827b5d5cb8748c8077bd835ae8c8bd97b8/src/Microsoft.OpenApi.OData.Reader/Vocabulary/Capabilities/UpdateRestrictionsType.cs#L17

As renaming the property would be breaking and would not match the spec name
https://github.com/oasis-tcs/odata-vocabularies/blob/a7850482cc83488d37c66c4e907f6a94406e1d05/vocabularies/Org.OData.Capabilities.V1.xml#L764C23-L764C33

irvinesunday: Thanks for the insightful heads up! I've updated as suggested.

andrueastman (approved):
👍🏼

Issue #585, opened 2024-09-30
Update tag names for action/function operations


irvinesunday:
Tag names for action/function paths need to be aligned the naming of all actions/functions tags to their corresponding _nav.
source.{navSource.entityTypeName/navProp.entityTypeName}_ name grouping. 

For example:

![image](https://github.com/user-attachments/assets/7f825253-e9ad-4dc2-9ad9-918b9ef7b628)

Operation: _admin.edge.internetExplorerMode.siteLists.browserSiteList.publish_ will move to the _admin.edge_ tag grouping.
 
In the OpenAPI...
 
The tag name for this action... 

![image](https://github.com/user-attachments/assets/ed280750-5831-4f7b-a68c-c915ccb49c9e)

needs to be updated to:

![image](https://github.com/user-attachments/assets/2ba4765d-946c-4e8b-9b3b-ca1c426de2c4)

PR #586, opened 2024-09-30
Updates tag names for actions/functions operations

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/585

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
@@ -133,10 +133,8 @@ namespace Microsoft.OpenApi.OData.Operation
         /// <inheritdoc/>
         protected override void SetTags(OpenApiOperation operation)
         {
-            string tagName = null;
-            var secondLastSegment = Path.Segments.Reverse().Skip(1).First();            
-            GenerateTagName(secondLastSegment, out tagName);            
-            OpenApiTag tag = new OpenApiTag
+            GenerateTagName(out string tagName);
+            OpenApiTag tag = new()
             {
                 Name = tagName,
             };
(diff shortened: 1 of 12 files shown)


src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
```diff
@@ -146,6 +146,31 @@ protected override void SetTags(OpenApiOperation operation)
...
         }
 
+        /// <summary>
+        /// Genrates the tag name for the operation.
+        /// </summary>
+        /// <param name="tagName">The generated tag name.</param>
+        /// <param name="skip">The number of segments to skip.</param>
+        private void GenerateTagName(out string tagName, int skip = 1)
+        {            
+            var targetSegment = Path.Segments.Reverse().Skip(skip).First();
```
andrueastman, line 156: Should we validate the length of the paths collection before skipping as `First()` could throw?

src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
```diff
@@ -146,6 +146,31 @@ protected override void SetTags(OpenApiOperation operation)
...
+            else if (targetSegment is ODataOperationSegment || targetSegment is ODataOperationImportSegment // Composable function
+                || targetSegment is ODataKeySegment) // Previous segmment could be a navigation property or a navigation source segment
+            {   
+                skip += 1;
+                GenerateTagName(out tagName, skip);
+            }
+            else // ODataNavigationSourceSegment
+            {
+                tagName = NavigationSource.Name + "." + NavigationSource.EntityType.Name;
+            }
```
andrueastman, line 171: Just a suggestion. This could be rewritten as a switch statement as below.

```suggestion
            switch (targetSegment)
            {
                case ODataNavigationPropertySegment:
                    tagName = EdmModelHelper.GenerateNavigationPropertyPathTagName(Path, Context);
                    break;
                case ODataOperationSegment:
                case ODataOperationImportSegment:
                // Previous segmment could be a navigation property or a navigation source segment
                case ODataKeySegment:
                    skip += 1;
                    GenerateTagName(out tagName, skip);
                    break;
                // ODataNavigationSourceSegment
                default:
                    tagName = NavigationSource.Name + "." + NavigationSource.EntityType.Name;
                    break;
            }
```

src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
irvinesunday: Operation paths will always have a navigation source as the 1st path segment, so `targetSegment` variable should ideally never be null. However, I can make the statement more resilient by swapping `First()` with `FirstOrDefault()` as follow:

`var targetSegment = Path.Segments.Reverse().Skip(skip).FirstOrDefault();`

Which will default in generating the default tag name:

```CSharp
else // ODataNavigationSourceSegment
{
    tagName = NavigationSource.Name + "." + NavigationSource.EntityType.Name;
}
```

src/Microsoft.OpenApi.OData.Reader/Operation/EdmOperationOperationHandler.cs
```diff
@@ -100,175 +100,204 @@ protected override void SetBasicInfo(OpenApiOperation operation)
...
+
+        /// <summary>
+        /// Genrates the tag name for the operation.
+        /// </summary>
+        /// <param name="tagName">The generated tag name.</param>
+        /// <param name="skip">The number of segments to skip.</param>
+        private void GenerateTagName(out string tagName, int skip = 1)
+        {            
+            var targetSegment = Path.Segments.Reverse().Skip(skip).FirstOrDefault();
+            targetSegment = null;
```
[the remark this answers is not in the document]
irvinesunday: This is an oversight. Shouldn't have been checked in.

andrueastman (approved):

Issue #582, opened 2024-09-20
Set request body to optional when all properties are optional


andrueastman:
Related to https://github.com/microsoftgraph/msgraph-sdk-java/pull/2163 and https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/2671#discussion_r1765012196

In a scenario where all the properties in the request body are optional such as an odata function with all parameters being optional, we should set the `requestBody.required = false` in the generated openApi description otherwise explictly set it to `true`.

This will unlock the generation of request body parameters in the SDKs as optional parameters in the builders to prevent breaking changes in scenarios where optional parametes are added.

andrueastman:
Once this is fixed and released, we should revert this PR. 
https://github.com/microsoftgraph/msgraph-metadata/pull/697

andrueastman:
Updated the description here. Should be Odata `action` not `function`. Functions do not have request bodies as the parameters are in the path.

irvinesunday:
Also, point of note is that, if `requestBody.required = false` the OAS does not emit the `required` property in the `requestBody` schema.

andrueastman:
According to the spec.
> Determines if the request body is required in the request. Defaults to false.

So what we should do is set it to true when required parameters are present...

irvinesunday:
Trying to retrieve the annotations in-lined in the Action parameters fails to retrieve the `Org.OData.Core.V1.OptionalParameter` annotation using the extensions from the Edm lib.
```CSharp
var annotations =  model.FindVocabularyAnnotationsIncludingInheritedAnnotations(parameter);
```

```XML
<Parameter Name="transitive" Type="Edm.Boolean" Nullable="false">
  <Annotation Term="Org.OData.Core.V1.OptionalParameter" />
</Parameter>
```

Checking with the OData team to check for its support.

zaaj:
So, @gavinbarron mentions in #3318 that this issue is related to the Restore-MgBetaDirectoryDeletedItem no longer having a -BodyParameter option, rendering all optional body parameters un-usable.

Gavin also mentioned MANY API calls all got optional BodyParameters added to them - does this mean ALL of the Powershell SDK auto-generated commands for those APIs are now broken in the sense that they have no -bodyparameter option?

PR #591, opened 2024-10-02
Sets request body required property to false when all action parameters are optional

irvinesunday:
Fixes https://github.com/microsoft/OpenAPI.NET.OData/issues/582

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiRequestBodyGenerator.cs
+++ b/src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiRequestBodyGenerator.cs
@@ -66,15 +66,17 @@ namespace Microsoft.OpenApi.OData.Generator
                 Properties = new Dictionary<string, OpenApiSchema>()
             };
 
+            bool allParamsNullable = true;
             foreach (var parameter in action.Parameters.Skip(skip))
             {
+                allParamsNullable = parameter.Type.IsNullable;
                 parametersSchema.Properties.Add(parameter.Name, context.CreateEdmTypeSchema(parameter.Type));
             }
 
-            OpenApiRequestBody requestBody = new OpenApiRequestBody
+            OpenApiRequestBody requestBody = new()
             {
                 Description = "Action parameters",
-                Required = true,
+                Required = !allParamsNullable,
                 Content = new Dictionary<string, OpenApiMediaType>()
             };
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiRequestBodyGenerator.cs
```diff
@@ -66,15 +66,17 @@ public static OpenApiRequestBody CreateRequestBody(this ODataContext context, IE
                 Properties = new Dictionary<string, OpenApiSchema>()
             };
 
+            bool allParamsNullable = true;
             foreach (var parameter in action.Parameters.Skip(skip))
             {
+                allParamsNullable = parameter.Type.IsNullable;
```
andrueastman, line 72: Won't this only pass the value of the last parameter that is evaluated in the loop? We probably need to do a bitwise comparison here

irvinesunday: Oh yes. You are right there

src/Microsoft.OpenApi.OData.Reader/Generator/OpenApiRequestBodyGenerator.cs
```diff
@@ -66,15 +66,17 @@ public static OpenApiRequestBody CreateRequestBody(this ODataContext context, IE
                 Properties = new Dictionary<string, OpenApiSchema>()
             };
 
+            bool allParamsNullable = true;
             foreach (var parameter in action.Parameters.Skip(skip))
             {
+                allParamsNullable &= parameter.Type.IsNullable;
```
andrueastman, line 72: One last question here. Should we be checking if the parameter is of type `IEdmOptionalParameter` instead? Is it possible for required parameter to be nullable?

irvinesunday: Yes indeed! We should. I will add a check for the annotation: `Org.OData.Core.V1.OptionalParameter`

andrueastman: I think for now let's only update the check to only cater for the annotation `Org.OData.Core.V1.OptionalParameter` as that's the scenario we want to cater for in https://github.com/microsoft/OpenAPI.NET.OData/issues/582. We can always come back and do that for nullable types.

irvinesunday: Okay

PR #628, opened 2024-12-24
Adds nullable to `double` schema conversions  (#589)

baywet:
port of #589 to the v1 branch


baywet:
this PR also ports #593

src/Microsoft.OpenApi.OData.Reader/Common/Constants.cs
```diff
@@ -168,6 +170,7 @@ internal static class Constants
         /// <summary>
         /// integer type
         /// </summary>
+        [Obsolete("integer is not a valid OpenAPI type. Use number instead.")]
         public static string IntegerType = "integer";
```
darrelmiller, line 174: I don't agree with this change.  Integer is a valid value for type in both OpenAPI and JSON Schema.

baywet: right, but as we own the conversion we can emit number directly, and formats where applicable. Part of the issue here is that we were emitting integer + a format, which is non-sensical.
I merely used the obsolete to track the references in the migration work, as well as to indicate this constant is going away in some near future since it's part of the public API surface.

darrelmiller: I wouldn't say "integer" + format is nonsensical, but I do agree that returning "number" + format is also valid and possibly less confusing.  There is the risk that some downstream code has made the assumption that number means decimal/float and ignores the format.  However, I think that is a risk we can take.

darrelmiller: I would recommend changing the comment to say "Integer is not a valid JSON type, therefore we are using number and format to fully specify all numeric formats".

baywet: what do you think of "Integer should not be used in combination with formats, therefore we're using number" or something along those lines?

andrueastman: Just checking here if the ask is to have this as a comment to the public property and not mark it as obsolete? 
Since the class that holds this property is `internal` it may be safe to even remove it if we no longer reference it as it won't be part of a public API(the class `Constants` isn't public.

baywet: good catch, removed the constant

andrueastman (approved):

PR #641, opened 2025-01-17
Adds action/function suffix to tag names for action/function operations

### LINKED ISSUE #640: Naming convention (``NavigationSource.NavigationEntity``) of tags targeting action/function operations causes incorrect Graph PowerShell module mappings
Changes made in #586 to address incorrect mappings in PowerShell as per the issue raised [here](https://github.com/microsoftgraph/msgraph-sdk-powershell/issues/2957), introduced mapping issues for paths that had ``Actions`` and ``Functions`` as suffixes in their tags.

To prevent breaking changes, there is need for those paths to remain in their [current modules](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/config/ModulesMapping.jsonc#L42).

timayabi2020:
Fixes #640 
This PR is a follow up to #586 and it includes "Actions" or "Functions" as suffixes to the tag. This will

- enable the [powershell mappings](https://github.com/microsoftgraph/msgraph-sdk-powershell/blob/dev/config/ModulesMapping.jsonc#L42) to be updated to target the paths more precisely to fix the incorrect mappings.
- enable the user functions/actions to still remain in their current modules by targeting operationIds like user.*.Actions.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
+++ b/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
@@ -33,6 +33,7 @@
 		- Further fix for generating unique operation ids for paths with composable overloaded functions where all functions in path are overloaded #594
 		- Further fix for generating unique operation ids for navigation property paths with composable overloaded functions #596
 		- Updates PUT operation id prefix from Update to Set #600
+    - Adds action/function suffix to tag names for actions/functions operations #641
 	</PackageReleaseNotes>
     <AssemblyName>Microsoft.OpenApi.OData.Reader</AssemblyName>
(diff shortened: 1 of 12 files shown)


baywet (changes requested):
Please add a unit test for that behaviour.
Also:
- main is the branch for v2.x (currently in preview)
- support/v1 is the branch for 1.X (GA, what our tooling uses)

You'll need to port the fix (cherry-pick, not merge) to benefit it in the generation pipeline

timayabi2020:
Closing this in favor of another PR that targets support/v1 branch

baywet:
@timayabi2020 you'll need to do both so the fix is also present in v2 when we release and there are no regressions. (unless it's already present in 2.X)

baywet (dismissed):

baywet:
@timayabi2020 I think some of the changes around the Id were missed here.

baywet (approved):

PR #645, opened 2025-01-22
Fix duplicate operation ids in overloads that are composable.

andrueastman:
Follow up to https://github.com/microsoft/OpenAPI.NET.OData/pull/642

Regression introduced due duplicate operation ids. The change doesn't seem like it was replicated in the v2 branch merge at https://github.com/microsoft/OpenAPI.NET.OData/pull/641. So no need of a second PR.

### DIFF
--- a/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
+++ b/src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
@@ -15,13 +15,13 @@
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <PackageId>Microsoft.OpenApi.OData</PackageId>
     <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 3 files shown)


baywet (approved):
Thanks for the fix, can you please also reflect the changes in v2?

andrueastman:
Just to confirm here @baywet, the v2 PR didn't end up with the same regression if you compare the diff of the two PRs,
https://github.com/microsoft/OpenAPI.NET.OData/pull/642 and https://github.com/microsoft/OpenAPI.NET.OData/pull/641.

Furthermore, the assertions on the head of the main(v2) branch looks correct/consistent with what we expect.
https://github.com/microsoft/OpenAPI.NET.OData/blob/b4bcdd23983de303d2b8cde4978c60e6e0cf6917/test/Microsoft.OpenAPI.OData.Reader.Tests/Operation/EdmFunctionOperationHandlerTests.cs#L381

baywet:
Thank you for double checking!

PR #668, opened 2025-03-18
chore: release 2.0.0-preview.10

MaggieKimani1:
Release-As: 2.0.0-preview.10

With the configuration of release please to enable tag-based releases, there's confidence that a tagged run is an intended release hence the need to remove the package version validation step.
Aligns with the OpenApi.NET library here: [microsoft/OpenAPI.NET@main/.azure-pipelines/ci-build.yml](https://github.com/microsoft/OpenAPI.NET/blob/main/.azure-pipelines/ci-build.yml?rgh-link-date=2025-03-18T14%3A01%3A29Z)


.azure-pipelines/ci-build.yml
```diff
@@ -148,14 +148,6 @@ extends:
               projects: src/Microsoft.OpenApi.OData.Reader/Microsoft.OpenAPI.OData.Reader.csproj
               arguments: '-o $(Build.ArtifactStagingDirectory) --configuration $(BuildConfiguration) --no-build --include-symbols --include-source /p:SymbolPackageFormat=snupkg'
 
-          - task: PowerShell@2
```
baywet, line 151: can you provide more context on why we're removing that validation step?

baywet (changes requested):

.azure-pipelines/ci-build.yml
andrueastman: I think in this case, it makes sense to drop the check and align with the openApi.Net library.
https://github.com/microsoft/OpenAPI.NET/blob/main/.azure-pipelines/ci-build.yml

Furthermore, since we now use release please and the deploy pipeline would run off of a tag and not every check-in on main, we have the confidence that the tagged run is an intended release (while previously, we would use this script as a filter for releasing). 

What do you think?

baywet: sure, I'm not pushing back, simply outlining the fact that providing this rationale in the PR description would have been great!
Also we should create a reflection of that PR for support/v1

baywet (approved):

PR #770, opened 2026-01-09
feat: add switch to use put as default update verb

gavinbarron:


baywet:
> SonarCloud build seems to be failing due to a config issue, should we address this in a different PR?

@adrian05-ms the analysis was configured in automatic mode for some reason which was conflicting with the CI. Not sure what happened but I flipped the setting back. Hopefully that solves it.

Issue #805, opened 2026-04-13
TryAddPath incorrectly resolves duplicate bound operation paths when derived types exist


gavinbarron:
## Bug Description

When processing bound operations (e.g., delta functions), `TryAddPath` in `Utils.cs` can produce incorrect return types in the generated OpenAPI specification. Specifically, when a base-type-bound operation (e.g., `directoryObject.delta()`) is processed before a derived-type-bound operation (e.g., `servicePrincipal.delta()`), the duplicate path resolution heuristic incorrectly keeps the base type''s operation.

## Reproduction

In the Microsoft Graph v1.0 CSDL:
- `/servicePrincipals/delta()` returns `Collection(graph.directoryObject)` instead of the correct `Collection(graph.servicePrincipal)`
- `/applications/delta()` correctly returns `Collection(graph.application)` (because it happens to be processed before the `directoryObject`-bound delta in document order)

## Root Cause

In `Utils.cs` `TryAddPath` (line ~373), when a duplicate path is detected:
1. The method checks if the **new** operation''s binding type has derived types
2. If yes, it assumes the **first** entry was more specific and keeps it
3. This is wrong when the first entry came from a **base type** and the new entry is the **exact match** for the entity set

## Trigger

The bug was latent until `servicePrincipal` gained derived types (`agentIdentity`, `agentIdentityBlueprintPrincipal`) in a March 2026 schema update. Before that, the else branch correctly replaced the base-type entry.

## Fix

Add an exact-match check: if the new operation''s binding type exactly matches the entity set''s type (`boundEntityType == operationEntityType`), always replace the existing entry regardless of derived types.

## Sub-issues

- #808 - OpenAPI 3.0 fix
- #806 - OpenAPI 3.1 fix
- #807 - OpenAPI 3.2 fix

gavinbarron:
FIx shipped

PR #809, opened 2026-04-13
fix: TryAddPath exact-match check for duplicate bound operation paths

### LINKED ISSUE #808: Fix TryAddPath duplicate bound operation resolution - OpenAPI 3.0
## Parent Issue

Tracked by #805

## Scope

Fix the `TryAddPath` duplicate path resolution bug for OpenAPI 3.0 output (support/v1 branch).

## Changes Required

1. **`Utils.cs`** - Add exact-match check in `TryAddPath`: when the new operation''s binding type exactly matches the entity set type, replace the existing entry
2. **`ODataPathProviderTests.cs`** - Add regression test reproducing the servicePrincipals/delta() scenario

gavinbarron:
## Summary

Fixes #808 (OpenAPI 3.0 track of #805)

When processing bound operations, `TryAddPath` can produce incorrect return types in the generated OpenAPI specification. This happens when a base-type-bound operation (e.g., `directoryObject.delta()`) is processed before a derived-type-bound operation (e.g., `servicePrincipal.delta()`), and the duplicate path resolution heuristic incorrectly keeps the base type''s operation.

## Problem

In the Microsoft Graph v1.0 CSDL, `/servicePrincipals/delta()` incorrectly returns `Collection(graph.directoryObject)` instead of the correct `Collection(graph.servicePrincipal)`.

**Root cause**: In `TryAddPath`, when a duplicate path is detected, the method checks if the new operation''s binding type has derived types. If yes, it keeps the first entry assuming it was more specific. This is wrong when the first entry came from a base type and the new entry is the exact match for the entity set.

## Fix

Added an exact-match check in `TryAddPath` (`Utils.cs`): if the operation''s binding type exactly matches the entity set type (`boundEntityType == operationEntityType`), always replace the existing entry regardless of derived types.

## Changes

- **`Utils.cs`**: Added exact-match guard before the existing derived-types heuristic
- **`ODataPathProviderTests.cs`**: Added regression test that reproduces the scenario with a minimal CSDL model

## Testing

All 1224 tests pass (1223 existing + 1 new regression test).


test/Microsoft.OpenAPI.OData.Reader.Tests/Edm/ODataPathProviderTests.cs
```diff
@@ -921,6 +922,76 @@ private static IEdmModel GetNavPropModel(string annotation)
...
+
+            var pathItem = doc.Paths["/servicePrincipals/NS.delta()"];
+            var getOp = pathItem.Operations[OperationType.Get];
+            Assert.NotNull(getOp);
+
+            // The response schema title should reference servicePrincipal, not directoryObject
+            var responseKey = getOp.Responses.Keys.First();
+            var response = getOp.Responses[responseKey];
+            var schema = response.Content["application/json"].Schema;
+            Assert.Contains("servicePrincipal", schema.Title);
```
MIchaelMainer, line 992:
```suggestion
            Assert.Contains("servicePrincipal", schema.Title);
            Assert.DoesNotContain("directoryObject", schema.Title);        
```

MIchaelMainer: nit: it is as important to have `serviceprincipal` as it is to not have `directoryobject`

MIchaelMainer (dismissed):

PR #845, opened 2026-07-17
Route NuGet restore through the CFS central package feed

gavinbarron:
Routes CI package restore through the authenticated `GraphDeveloperExperiences_Public` Azure Artifacts feed (network isolation / CFS compliance).

- Adds `NuGetAuthenticate@1` and generates a `nuget.config` with `<clear/>` + the central feed as the only source.

Note: requires a pipeline run to validate against the feed.

### DIFF
--- a/.azure-pipelines/ci-build.yml
+++ b/.azure-pipelines/ci-build.yml
@@ -74,6 +74,20 @@ extends:
               checkLatest: true
 
           # Build the Product project
+          - task: NuGetAuthenticate@1
+            displayName: 'Authenticate to Azure Artifacts'
+
+          - pwsh: |
+              @"
+              <?xml version="1.0" encoding="utf-8"?>
+              <configuration>
+                <packageSources>
+                  <clear />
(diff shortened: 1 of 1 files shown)


.azure-pipelines/ci-build.yml
```diff
@@ -74,6 +74,20 @@ extends:
               checkLatest: true
 
           # Build the Product project
+          - task: NuGetAuthenticate@1
```
baywet, line 77: Shouldn't this be after setting the feed? So it can use that information to auth.

gavinbarron: weirdly, not for NuGet, it uses a different style of auth from NPM 
Here's the key part of the output from the CI run for this step on this PR

```pwsh
Setting up the credential provider to use the identity 'Graph Developer Experiences Build Service (microsoftgraph)' for feeds in your organization/collection starting with:
  https://microsoftgraph.pkgs.visualstudio.com/
  https://pkgs.dev.azure.com/microsoftgraph/

```

baywet (approved):

